PR #1, opened 2022-06-10
Add support for sparse LQ dynamic optimization

dlcole3:
Added functions to `DynamicNLPModels.jl` to enable building a `QuadraticModel` from a Q, R, A, and B matrix with optional x0, Qf, and lower/upper bounds on state and input variables. 

3 functions were added:
- `build_H`: builds a sparse matrix from the Q and R (and optional Qf) matrices such that $z^T H z = \Sigma_{i=1}^N(x_i^T Q x_i) + \Sigma_{i=1}^N(u_i^T R u_i) + x_N^T Qf x_N$ (where $Qf$ defaults to $Q$ if not defined). This is the format `QuadraticModels.jl` expects
- `build_A`: builds a sparse matrix that is equivalent to $x_(t+1) = Ax_t + Bu_t,$ for $t = 1, ...., N-1$. The new matrix has the form $ 0 \le Az \le 0$, which is the format `QuadraticModels.jl` expects.
- `get_QM`: Constructs a `QuadraticModel` from the Q, R, A, and B matrices along with the number of time steps, nt. Returns a `QuadraticModel`. 

Also added scripts to `test/` that includes a script building an equivalent `JuMP.jl` model of the above quadratic programs and tests that the solution fo the `QuadraticModel` is the same as the `JuMP.jl` model. The `runtests.jl` file runs the `JuMP.jl` model with `Ipopt` and compares it to the `QuadraticModel` solved with `Ipopt` and with `MadNLP`


src/DynamicNLPModels.jl
```diff
@@ -29,11 +33,130 @@ function LQDynData(
...
+
 end
 
 function QPData(dnlp::LQDynData{VT,MT}) where {VT,MT}
-    # TODO
+    
+end
+
+""" 
+Build the (sparse) H matrix for quadratic models from Q and R matrices
```
sshin23: Checkout the guidelines for docstring https://docs.julialang.org/en/v1/manual/documentation/

The docstring should include what are the arguments for the function. E.g., `?push!` gives me:

```
  push!(collection, items...) -> collection


  Insert one or more items in collection. If collection is an ordered
  container, the items are inserted at the end (in the given order).

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> push!([1, 2, 3], 4, 5, 6)
  6-element Vector{Int64}:
   1
   2
   3
   4
   5
   6


  If collection is ordered, use append! to add all the elements of another
  collection to it. The result of the preceding example is equivalent to
  append!([1, 2, 3], [4, 5, 6]). For AbstractSet objects, union! can be used
  instead.

  See sizehint! for notes about the performance model.

  See also pushfirst!.
```

src/DynamicNLPModels.jl
```diff
@@ -29,11 +33,130 @@ function LQDynData(
...
+Objective function is 1/2 z^T H z = 1/2 sum(x^T Q x for i in 1:T) + 1/2 sum(u^T R u for i in 1:(T-1))
+"""
+function build_H(
+    Q, R, nt;
+    Qf = [])
+    if size(Qf)[1] == 0
+        Qf = copy(Q)
+    end
+
+    ns = size(Q)[1]
```
sshin23: `size(Q,1)`

src/DynamicNLPModels.jl
```diff
@@ -29,11 +33,130 @@ function LQDynData(
...
+    return H
+end
+
+
+"""
+Build the (sparse) A matrix for quadratic models from the Ac and B matrices
+where 0 <= Az <= 0 for x_t+1 = Ac* x_t + B* u_t
+"""
+
+function build_A(Ac,B, nt)
```
sshin23: Using `A` for jacobian is confusing. Can you use `J` for jacobian and `A` for the system matrix?

src/DynamicNLPModels.jl
```diff
@@ -29,11 +33,130 @@ function LQDynData(
...
+    ns = size(Ac)[2]
+    nr = size(B)[2]
```
sshin23: `size(Ac,2)`

src/DynamicNLPModels.jl
```diff
@@ -29,11 +33,130 @@ function LQDynData(
...
+
+ 
+"""
+Get the QuadraticModels.jl QuadraticModel from the Q, R, A, and B matrices
+nt is the number of time steps
+"""
+function get_QM(
+    Q, R, A, B, nt;
+    c    = zeros(nt*size(Q)[1] + nt*size(R)[1]),
+    lvar = fill(-Inf, (nt*size(Q)[1] + nt*size(R)[1])),
```
sshin23: instead of `lvar` and `uvar`, we should take `xl` and `xu` for `xl <= x <= xu`

test/JuMP_sparse_linear_mpc.jl
```diff
@@ -0,0 +1,76 @@
+using Ipopt, JuMP, Random, LinearAlgebra
```
sshin23: We need to specify the dependency for test in `Project.toml`. Check https://github.com/MadNLP/MadNLP.jl/blob/master/Project.toml#L27-L34

test/JuMP_sparse_linear_mpc.jl
```diff
@@ -0,0 +1,76 @@
...
+            for i in NU
+                @constraint(model, u[i,j] <= uvar[ns*nt + (j-1)*nu + i])
+            end
+
+        end
+    end
+
+    if length(x0) != 0
+        for i in NS
+            @constraint(model, x[i,1] == x0[i])
```
sshin23: to be consistent with what we have for QMs, let's replace this to bound constraints for variables

test/runtests.jl
```diff
@@ -1,3 +1,135 @@
...
+ipopt_ulbx0   = ipopt(qpulbx0)
+madnlp_ulbx0  = madnlp(qpulbx0)
+
+# Test results
+@test abs(objective_value(mulbx0) - ipopt_ulbx0.objective)  < 1e-7
+@test abs(objective_value(mulbx0) - madnlp_ulbx0.objective) < 1e-7
+@test abs(madnlp_ulbx0.objective  - ipopt_ulbx0.objective) < 1e-7
+
+
+
```
sshin23: Let's remove 104-end. In principle, tests shouldn't fail as long as we have a correct implementation, though we want it to cover as many edge cases as possible. What we have here is not an edge case, since we may actually have an infeasible problem, due to the randomness. If we want to test that our package correctly identifies the infeasibility, we need to *intentionally* put an infeasible problem and see if the solver actually exits with infeasibility.

sshin23 (changes requested):
Not a bad PR for a first PR! Could you work on the requested changes @dlcole3?

test/JuMP_sparse_linear_mpc.jl
dlcole3: In other words, I should not treat $x_0$ as a variable, correct? Right now, I have been treating it as a variable, but it makes sense that I should remove that.

sshin23: ah, sorry. what I meant is "fixing" variables https://jump.dev/JuMP.jl/stable/manual/variables/

src/DynamicNLPModels.jl
dlcole3: In this case, xl and xu are  length nx (they are NOT length nx*N) correct?

sshin23: correct. And we do the same for `u`

sshin23:
and another minor comment. I find it confusing to use `x` as the symbol for state variables, because `x` is used for primal variable for NLP. I suggest changing the symbols for state variables to `s`

Project.toml
```diff
@@ -16,6 +16,12 @@ julia = "^1.6.0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+MadNLP = "2621e9c9-9eb4-46b1-8089-e8c72242dfb6"
+NLPModelsIpopt = "f4238b75-b362-5c4c-b852-0801c9a21d71"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
```
sshin23: LinearAlgebra is already in [deps], so no need to list it in [extras]

src/DynamicNLPModels.jl
```diff
@@ -2,6 +2,10 @@ module DynamicNLPModels
 
 import NLPModels
 import QuadraticModels
+import LinearAlgebra
+import SparseArrays
+
+export build_H, build_J, get_QM
```
sshin23: no need to export `build_H` and `build_J`, and rename it to `_build_H` and `_build_J`

sshin23: see https://stackoverflow.com/questions/53687998/function-name-with-a-leading-underscore

test/JuMP_sparse_linear_mpc.jl
```diff
@@ -0,0 +1,231 @@
+using Ipopt, JuMP, Random, LinearAlgebra
+
+
+
+"""
```
sshin23: it's good to have the docstring for this, but not particularly necessary since we are using this function just for testing. Next time don't spend too much time on docstrings for testing

test/JuMP_sparse_linear_mpc.jl
```diff
@@ -0,0 +1,231 @@
...
+
+    # Give objective function as xT Q x + uT R u where x is summed over T and u is summed over T-1
+    @objective(model,Min,  sum( 1/2 * Q[s1, s2]*(s[s1,t])*(s[s2,t]) for s1 in NS, s2 in NS, t in 1:(N-1)) + 
+            sum( 1/2 * R[u1,u2] * u[u1, t] * u[u2,t] for t in 1:(N-1) , u1 in NU, u2 in NU) + 
+            sum( 1/2 * Qf[s1,s2] * s[s1,N] * s[s2, N]  for s1 in NS, s2 in NS))
+
+    # return model
+    return model
+end
+#=
```
sshin23: remove the comment

test/JuMP_sparse_linear_mpc.jl
```diff
@@ -0,0 +1,231 @@
+using Ipopt, JuMP, Random, LinearAlgebra
```
sshin23: some duplicate `using` here. Ideally, we want to `using` all package at the beginning of `runtests.jl` script

sshin23: and possibly rename the script to `sparse_lq_test.jl`

sshin23 (changes requested):
Some further comments

test/JuMP_sparse_linear_mpc.jl
dlcole3: Does this apply to internal functions that are not being exported as well, such as `_build_J` or `_build_H`?

Project.toml
```diff
@@ -16,6 +16,11 @@ julia = "^1.6.0"
...
+MadNLP = "2621e9c9-9eb4-46b1-8089-e8c72242dfb6"
+NLPModelsIpopt = "f4238b75-b362-5c4c-b852-0801c9a21d71"
+JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
+Ipopt = "b6b21f68-93f8-5de0-b562-5493be1d77c9"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [targets]
-test = ["Test"]
\ No newline at end of file
+test = ["Test", "MadNLP", "NLPModelsIpopt", "JuMP", "Ipopt", "Random"]
```
sshin23, line 26: LinearAlgebra here

sshin23 (changes requested):
minor comment

Project.toml
dlcole3: I originally had that in there, but I the test failed (see [link](https://github.com/MadNLP/DynamicNLPModels.jl/runs/6838304172?check_suite_focus=true)). Was I missing something on that original commit? This is from the test run after commit [e9f13f7](https://github.com/MadNLP/DynamicNLPModels.jl/pull/1/commits/e9f13f7fc112a83bdf932314a7441f84124202d9)

sshin23: ah, yes, you're right. no need for the change

sshin23 (approved):
approve all the changes

PR #7, opened 2022-06-13
Defining data structure for LQ dynamic optimization

### LINKED ISSUE #2: Defining data structure for LQ dynamic optimization
We should define the following data structure for LQ dynamic optimization
```julia
abstract type AbstractDynamicData{T,S} end

struct LQDynamicData{T,S,M} <: AbstractDynamicData{T,S}
s0::S
A::M
B::M
Q::M
R::M
Qf::M
# etc
end

abstract type AbstractDynamicModel{T,S} <: QuadraticModels.AbstractQuadraticModel{T, S} end

mutable struct LQDynamicModel{T, S, M1, M2, M3} <:  AbstractDynamicModel{T,S} 
  meta::NLPModelMeta{T, S}
  counters::Counters
  data::QPData{T, S, M1, M2}
  dynamic_data::LQDynamicData{T,S,M3}
  condense::Bool
end
```

and 
- define a constructor for `LQDynamicData`
- replace the current `build_QM` with `LQDynamicModel(::LQDynamicData{T,S,M}; condense = false) where {T,S,M}`
- define dynamic optimization-specific APIs such as `get_s0` and `set_s0!`

dlcole3:
This PR provides 3 things in response to issue "Defining data structure for LQ dynamic optimization".
- It defines the struct `LQDynamicData` and provides a constructor for this struct. There are two constructor options:
    - It can build the `LQDynamicData` by passing the `Q`, `R`, `A`, `B` matrices, `s0` vector, and the `N` time steps with optional arguments for `sl`, `su`, `ul`, and `uu`.
    - It can build the `LQDynamicData` by passing `ns`, `nu`, and `N`. It initializes empty arrays for `Q`, `R`, `A`, and `B` and sets `s0` to a zero vector and defaults to infinite bounds for states and inputs. These values can then be adjusted with the API functions
- It defines a new `LQDynamicModel(::LQDynamicData{T,S,M}; condense = false) where {T,S,M}` in the place of `build_QM`. 
- It defines functions for APIs, so that attributes of `LQDynamicData` can be modified or queried. The only exception is that `ns`, `nu`, and `N` cannot be modified.


src/DynamicNLPModels.jl
```diff
@@ -4,42 +4,597 @@ import NLPModels
...
 import LinearAlgebra
 import SparseArrays
+import SparseArrays: SparseMatrixCSC
 
-export get_QM
+export get_QM, LQDynamicData, LQDynamicModel,
+get_s0, set_s0!, get_Q, set_Q!, get_R, set_R!, get_Qf,
+set_Qf!, get_A, set_A!, get_B, set_B!, get_sl, set_sl!, 
+get_su, set_su!, get_ul, set_ul!, get_uu, set_uu!,
+jac_structure!, hess_structure!, jac_coord!, hess_coord!
```
sshin23: no need to export `jac_structure!, hess_structure!, jac_coord!, hess_coord!`

src/DynamicNLPModels.jl
```diff
@@ -4,42 +4,597 @@ import NLPModels
...
+        error("x0 is not within the given upper and lower bounds")
+    end
+
+    ns= size(Q,1)
+    nu= size(R,1)
+
+
+    LQDynamicData{T,S,M}(
+        s0, A, B, Q, R, N,
+
```
sshin23: the empty line should be removed

src/DynamicNLPModels.jl
```diff
@@ -4,42 +4,597 @@ import NLPModels
...
+- `ul = fill(-Inf, nu)`: vector of lower bounds on input variables
+- `uu = fill(Inf, nu)` : vector of upper bounds on input variables
+"""
+function LQDynamicData(
+    s0::S,
+    A::M,
+    B::M,
+    Q::M,
+    R::M,
+    N::Int;
```
sshin23: I'd remove `::Int` here just to make the constructor more flexible

src/DynamicNLPModels.jl
```diff
@@ -4,42 +4,597 @@ import NLPModels
...
+---
+
+Initializes empty `Q`, `R`, `A`, and `B` (sparse) arrays and empty `s0`, `sl`, `su`, `ul`, and `uu` vectors
+
+Inputs:
+- `ns`: number of states
+- `nu`: number of inputs
+- `N` : number of time points
+"""
+function LQDynamicData(ns::Int,
```
sshin23: this seems to be unnecessary

src/DynamicNLPModels.jl
```diff
@@ -4,42 +4,597 @@ import NLPModels
...
+    su = fill(Inf, ns)
+    ul = fill(-Inf, nu)
+    uu = fill(Inf, nu)
+
+    LQDynamicData(s0, A, B, Q, R, N; Qf = Qf, sl = sl, su = su, ul = ul, uu = uu)
+
+end
+
+
+
```
sshin23: for these functions, see if you can use @eval to make it more concise. See:
https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/main/src/nlp/tools.jl#L5-L17

src/DynamicNLPModels.jl
```diff
@@ -4,42 +4,597 @@ import NLPModels
...
+        c0, 
+        c,
+        H,
+        J
+        ),
+        dnlp,
+        condense
     )
 
 end
```
sshin23, line 247: Let's define another constructor:
```julia
function LQDynamicModel(
    s0::S,
    A::M,
    B::M,
    Q::M,
    R::M,
    N;
    Qf::M = Q, 
    sl::S = (similar(s0) .= -Inf),
    su::S = (similar(s0) .=  Inf),
    ul::S = (similar(s0,size(R,1)) .= -Inf),
    uu::S = (similar(s0,size(R,1)) .=  Inf)
    ) where {T,S <: AbstractVector{T},M <: AbstractMatrix{T}}
```

src/DynamicNLPModels.jl
```diff
@@ -4,42 +4,597 @@ import NLPModels
...
 
-export get_QM
+export get_QM, LQDynamicData, LQDynamicModel,
+get_s0, set_s0!, get_Q, set_Q!, get_R, set_R!, get_Qf,
+set_Qf!, get_A, set_A!, get_B, set_B!, get_sl, set_sl!, 
+get_su, set_su!, get_ul, set_ul!, get_uu, set_uu!,
+jac_structure!, hess_structure!, jac_coord!, hess_coord!
 
-mutable struct LQDynData{VT,MT}
+abstract type AbstractDynamicData{T,S} end
```
sshin23: I'd call it AbstractLQDynamicData

sshin23 (changes requested):
Nice job, @dlcole3!

Only have a couple of minor comments

src/DynamicNLPModels.jl
sshin23: this will make your job of maintaining the code much easier ;)

PR #8, opened 2022-06-14
Hessian bug fix

### LINKED ISSUE #4: Hessian bug fix
Normally, MadNLP shouldn't regularize the problem for convex QPs, but it is regularizing this problem. This is an indication of a potential bug. This should be root-caused and fixed
https://github.com/MadNLP/DynamicNLPModels.jl/runs/6838395675?check_suite_focus=true#step:5:656

dlcole3:
Resolved the bug that resulted in regularizing a convex QP within `MadNLP.jl`. 

`LQDynamicModel` included a $u_N$ variable that was not needed within the optimization problem. Removed this variable by removing empty row/column in Hessian and empty column in Jacobian and extra lines in `lvar` and `uvar`. 

Also updated `build_QP_JuMP_model` function within `sparse_lq_test.jl` to remove the $u_N$ variable.


examples/JuMP_sparse_linear_mpc.jl
```diff
@@ -0,0 +1,79 @@
+using Ipopt, JuMP, Random, LinearAlgebra
```
sshin23: these files should be removed

sshin23 (approved):
approve the changes, but remove the old files in `examples`

src/DynamicNLPModels.jl
```diff
@@ -562,6 +562,55 @@ function _build_H(
 end
```
sshin23: shouldn't appear yet

sshin23 (changes requested):

src/DynamicNLPModels.jl
dlcole3: Sorry, I thought I had done my branch correctly to start the dense formulation, but that is not the case. I will remove that and push again

sshin23 (approved):

PR #15, opened 2022-06-16
Support for `gl <= Ex + Fu <= gu` constraints added for sparse and dense formats

### LINKED ISSUE #5: Add support for `gl <= Ex+Fu <= gu` constraints
Currently DynamicNLPModels does not support `gl <= Ex+Fu <= gu` type constraints. We should support this additional constraint, once we have the basic sparse/dense functionality NLP working

dlcole3:
Updated `LQDynamicData` to store attributes for E, F, gl, and gu data. Updated the `LQDynamicModel` constructor to take E, F, gl, and gu data and build it into the model for both sparse and condensed formats. Added a `_build_EF` function for the sparse format. Updated the `sparse_LQ_test.jl` file to solve the `JuMP.jl` model with the E and F data. Added tests to `runtests.jl` to test the problem with new E and F matrices.


src/DynamicNLPModels.jl
```diff
@@ -51,45 +55,48 @@ struct LQDynamicData{T,S,M} <: AbstractLQDynData{T,S}
...
 - `R` : objective function matrix for system inputs from 1:(N-1)
 - `N` : number of time steps
-
 The following attributes of the `LQDynamicData` type are detected automatically from the length of s0 and size of R
 - `ns`: number of state variables
 - `nu`: number of input varaibles
-
 The following keyward arguments are also accepted
-- `Qf`: objective function matrix for system state at time N; defaults to Q
+- `Qf = Q`: objective function matrix for system state at time N
```
frapac: Maybe specify `Qf` dimension?

src/DynamicNLPModels.jl
```diff
@@ -157,27 +190,25 @@ end
...
                 s_0 = s0
 ```
 ---
 
 If `condense=false`, data is converted to the form 
+
 ```math
     minimize    \\frac{1}{2} z^T H z 
-    subject to  0 \\le Jz \\le 0
+    subject to  lcon \\le Jz \\le ucon
```
frapac, line 209: Ok for now. But we may want to update the docstring later for a better display in LaTeX

frapac (approved):
It overall looks good to me

src/DynamicNLPModels.jl
```diff
@@ -205,39 +235,49 @@ function LQDynamicModel(dnlp::LQDynamicData{T,S,M}; condense = false) where {T,S
...
     uu = dnlp.uu
+    gl = dnlp.gl
+    gu = dnlp.gu
 
 
     if condense == false
-        H = _build_H(Q, R, N; Qf=Qf)
-        J = _build_J(A, B, N)
+        H  = _build_H(Q, R, N; Qf=Qf)
+        J  = _build_J(A, B, N)
```
sshin23: Instead of `J` and `EF`, let's call it `J1` and `J2` The current naming can potentially cause a confusion

src/DynamicNLPModels.jl
```diff
@@ -205,39 +235,49 @@ function LQDynamicModel(dnlp::LQDynamicData{T,S,M}; condense = false) where {T,S
...
 
     sl = dnlp.sl
     su = dnlp.su
     ul = dnlp.ul
     uu = dnlp.uu
+    gl = dnlp.gl
+    gu = dnlp.gu
 
 
     if condense == false
```
sshin23: As this constructor gets complicated, let's separate it into

```julia
function _build_sparse_lq_dyanmic_model
function _build_dense_lq_dyanmic_model
```

src/DynamicNLPModels.jl
```diff
@@ -257,18 +297,28 @@ function LQDynamicModel(dnlp::LQDynamicData{T,S,M}; condense = false) where {T,S
...
-            Jcon[i + ns, i] = 1.0
-        end
-        
         d[1:ns] .= .-sl
         d[(ns+1):(2*ns)] .= su
 
-        J, ucon = _build_G(block_A, block_B, Jcon, E, d, s0, N)
+
+        E_bounds = zeros(ns * 2, ns)
+        F_bounds = zeros(ns * 2, nu)
```
sshin23: Rather than treating them as part of `E` and `F` block, you can explicitly derive the corresponding Jacobian block and put it there. The current approach causes some unnecessary allocations

src/DynamicNLPModels.jl
```diff
@@ -417,33 +472,33 @@ function _build_condensed_blocks(
...
   
-    G = similar(block_E)
+    G = zeros(size(block_F))
   
     As0  = zeros(size(block_A, 1), 1)
-    JAs0 = zeros(size(block_J, 1), 1)
+    EAs0 = zeros(size(block_E, 1), 1)
 
-    LinearAlgebra.axpy!(1, block_E, LinearAlgebra.mul!(G, block_J, block_B))
+    LinearAlgebra.axpy!(1, block_F, LinearAlgebra.mul!(G, block_E, block_B))
```
sshin23: Let's separate `axpy!` and `mul!`

src/DynamicNLPModels.jl
```diff
@@ -257,18 +297,28 @@ function LQDynamicModel(dnlp::LQDynamicData{T,S,M}; condense = false) where {T,S
...
+
+        for i in 1:ns
+            E_bounds[i,i] = -1.0
+            E_bounds[i + ns, i] = 1.0
+        end
+
+        E_bounds = vcat(E_bounds, E)
+        F_bounds = vcat(F_bounds, F)
+        d        = vcat(d, gu)
+
```
sshin23, line 433: Double-sided inequalities should be used; in this way, the NLP solver needs to handle x2 more constraints

sshin23 (changes requested):
nice work @dlcole3. A few changes are needed

src/DynamicNLPModels.jl
```diff
@@ -205,39 +235,49 @@ function LQDynamicModel(dnlp::LQDynamicData{T,S,M}; condense = false) where {T,S
...
-        nnzh = length(H.rowval)
-        ncon = size(J, 1)
 
         c  = zeros(nvar)
 
-        lvar = copy(s0)
-        uvar = copy(s0)
-        ucon  = zeros(ncon)
-        lcon  = zeros(ncon)
+        lvar  = copy(s0)
```
sshin23: try to minimize the use of `vcat`. A better way is to first create an array of proper size, and set the entry of the array

src/DynamicNLPModels.jl
```diff
@@ -417,33 +472,33 @@ function _build_condensed_blocks(
     return (H = BQB, c = vec(h), c0 = c0, block_A = block_A, block_B = block_B, block_Q = block_Q, block_R = block_R)
 end
 
-function _build_G(block_A, block_B, J, E, d,s0, N)
+function _build_G(block_A, block_B, E, F, d,s0, N)
```
sshin23: it would be more intuitive to accept `block_E` and `block_F`

src/DynamicNLPModels.jl
```diff
@@ -257,18 +297,28 @@ function LQDynamicModel(dnlp::LQDynamicData{T,S,M}; condense = false) where {T,S
...
         d[(ns+1):(2*ns)] .= su
 
-        J, ucon = _build_G(block_A, block_B, Jcon, E, d, s0, N)
+
+        E_bounds = zeros(ns * 2, ns)
+        F_bounds = zeros(ns * 2, nu)
+
+        for i in 1:ns
+            E_bounds[i,i] = -1.0
+            E_bounds[i + ns, i] = 1.0
```
sshin23: don't forget the edge case where some entries of `s` are unbounded

src/DynamicNLPModels.jl
```diff
@@ -652,9 +707,9 @@ function _build_H(
     end
 
     ns = size(Q, 1)
-    nr = size(R, 1)
+    nu = size(R, 1)
 
-    H = SparseArrays.sparse([],[],Float64[],(ns * (N + 1) + nr * N), (ns * (N+1) + nr * N))
+    H = SparseArrays.sparse([],[],Float64[],(ns * (N + 1) + nu * N), (ns * (N+1) + nu * N))
```
sshin23: Try to not assume that the float precision is `Float64`. It is better to infer it from one of the function arguments

src/DynamicNLPModels.jl
```diff
@@ -732,6 +787,29 @@ function _build_J(A,B, N)
     return J
 end
 
+function _build_EF(E, F, N)
```
sshin23: how about rename these functions with indication of whether it's for sparse formulation or dense formulation? Say, `_build_dense_J2`, etc.

sshin23 (commented):
some further comments

sshin23 (approved):

sshin23:
excellent work, @dlcole3

PR #18, opened 2022-06-17
Added support for off-diagonal S elements

### LINKED ISSUE #12: Add support for off-diagonal objective `S`
Add support for off-diagonal objective `S`. See:
https://www.sciencedirect.com/science/article/pii/S0005109812001069

dlcole3:
Added support to `LQDynamicData` and `LQDynamicModel` to handle non diagonal `S` matrix in the objective function. `LQDynamicData` now has an attribute for `S`, and `LQDynamicModel` now incorporates `S` into the construction of `H` for both sparse and condensed forms. Also updated `sparse_lq_test.jl` so that the JuMP formulation includes the `S` matrix and added a new test for the addition of the `S` matrix. Also changed type variable in source code so that it uses `V` in place of `S` (e.g., `LQDynamicData{T, S, M}` is now `LQDynamicData{T, V, M}`.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -15,7 +15,7 @@ abstract type AbstractLQDynData{T,V} end
 A struct to represent the features of the optimization problem 
 
 ```math
-    minimize    \\frac{1}{2} \\sum_{i = 0}^{N-1}(s_i^T Q s_i + u_i^T R u_i) + \\frac{1}{2} s_N^T Qf s_N
+    minimize    \\frac{1}{2} \\sum_{i = 0}^{N-1}(s_i^T Q s_i + 2 u_i^T S^T x_i + u_i^T R u_i) + \\frac{1}{2} s_N^T Qf s_N
     subject to  s_{i+1} = A s_i + B u_i  for i=0, 1, ..., N-1
                 gl \\le E s_i + F u_i \\le gu for i = 0, 1, ..., N-1
                 sl \\le s \\le su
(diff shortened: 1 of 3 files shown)


src/DynamicNLPModels.jl
```diff
@@ -576,6 +589,29 @@ function _build_condensed_blocks(
   
     c0 = h0[1,1] / 2
   
+    if S != zeros(size(Q, 1), size(R, 1))
```
sshin23: `zeros(size(Q, 1), size(R, 1))` creates an array. Use `iszero` instead

src/DynamicNLPModels.jl
```diff
@@ -576,6 +589,29 @@ function _build_condensed_blocks(
...
+
+        BTS      = zeros(nu * N, nu * N)
+        STB      = zeros(nu * N, nu * N)
+        s0T_AT_S = zeros(1, nu * N)
+
+        LinearAlgebra.mul!(BTS, transpose(block_B), block_S)
+        LinearAlgebra.mul!(STB, transpose(block_S), block_B)
+
+        LinearAlgebra.axpy!(1, BTS, BQB)
+        LinearAlgebra.axpy!(1, STB, BQB)
```
sshin23: is `STB` a transpose of `BTS`? If so, no need to do the computation twice, but just use `BTS'` in place of `STB`

src/DynamicNLPModels.jl
```diff
@@ -493,7 +504,8 @@ end
 
 function _build_condensed_blocks(
     s0, Q, R, A, B, N;
-    Qf = Q)
+    Qf = Q, 
+    S = zeros(size(Q, 1), size(R, 1)))
```
sshin23: consider initializing `S=nothing`. In that way, we can avoid unnecessarily allocating memory

sshin23 (approved):
looks good @dlcole3, please merge the PR once the comments are addressed

PR #22, opened 2022-06-20
Added support for `K` state elimination

### LINKED ISSUE #11: Add support for `K` for state elimination
Add support for `K` for state elimination. See:
https://www.sciencedirect.com/science/article/pii/S0005109812001069

dlcole3:
Added support for both sparse and condensed formulations of the `K` state elimination matrix. 
 - Added the `K` matrix as an attirbute of the `LQDynamicData`
 - Added function for building the `J1` matrix in the sparse format when `K` is defined so that it includes the `K` constraints
 - Added two new functions for building the condensed blocks and the condensed `G` block for the condensed formulation since the algebra is signficantly more complex with `K` defined. 
 - Separated `_build_condensed_lq_dynamic_model` into two functions, one for when `K` is defined and one when it is not. These functions are called separately in `LQDynamicModel`. 
 - Added new tests for the `K` matrix formulation. Added tests for when `K` is defined with bounds, when `K` and `S` are defined with bounds, when `K` is defined without bounds, and when `K` and `S` are defined and not all `u` variables are bounded. 
 - Updated the `sparse_lq_test.jl` file so that it forms the `JuMP.jl` model with the `K` matrix.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -59,6 +59,7 @@ struct LQDynamicData{T,V,M} <: AbstractLQDynData{T,V}
     nu::Int
     E::M
     F::M
+    K::Union{M, Nothing}
 
     sl::V
     su::V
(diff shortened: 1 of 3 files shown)


dlcole3:
Code needs a lot of revision. Need to use multiple dispatch. I am closing this PR and will open a new one when I have revised the code.

PR #23, opened 2022-06-21
Added support for `K` state elimination

dlcole3:
Added both sparse and condensed support for `K` state elimination. 
 - Added attribute to `LQDynamicData` with new type for `K::MK`. This allows for multiple dispatch of different functions.
 - Redefined `_build_condensed_lq_dynamic_model` and `_build_sparse_lq_dynamic_model` for different types of `K`. The matrix `K` influences problem type and constraint and variable bounds. 
 - Transformed `A` and `E` matrices within `_build_sparse_lq_dynamic_model` for when `K` is defined. 
 - Added `_build_sparse_J3` function that is called within the `_build_sparse_lq_dynamic_model` when `K <: AbstractMatrix`. This converts $u$ variable bounds to algebraic constriants for $v$
 - Redefined `_build_condensed_blocks` to just build the block matrices of the condensed problems
 - Defined `_build_condensed_H_blocks` that returns `H`, `h`, and `h0`. Defined two forms of the function depending on `K` type
 - Defined `_build_condensed_G_blocks` that returns `J`, `lcon`, and `ucon`. 
 - Updated `sparse_lq_test.jl` to include a `K` matrix
 - Updated `runtests.jl` to test the problem with a `K` matrix, with a `K` matrix and `S` matrix, with a `K` matrix and `S` matrix and partial bounds on $u$, and with the `K` matrix and no bounds.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -10,13 +10,14 @@ export LQDynamicData, LQDynamicModel, _build_condensed_blocks
 
 abstract type AbstractLQDynData{T,V} end
 """
-    LQDynamicData{T,V,M} <: AbstractLQDynData{T,V}
+    LQDynamicData{T,V,M,MK} <: AbstractLQDynData{T,V}
 
 A struct to represent the features of the optimization problem 
 
 ```math
     minimize    \\frac{1}{2} \\sum_{i = 0}^{N-1}(s_i^T Q s_i + 2 u_i^T S^T x_i + u_i^T R u_i) + \\frac{1}{2} s_N^T Qf s_N
     subject to  s_{i+1} = A s_i + B u_i  for i=0, 1, ..., N-1
+                u_i = Kx_i + v_i  \\forall i = 0, 1, ..., N - 1
                 gl \\le E s_i + F u_i \\le gu for i = 0, 1, ..., N-1
                 sl \\le s \\le su
                 ul \\le u \\le uu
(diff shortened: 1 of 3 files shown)


src/DynamicNLPModels.jl
```diff
@@ -94,6 +98,7 @@ The following keyward arguments are also accepted
 - `S  = nothing`: objective function matrix for system state and inputs
 - `E  = zeros(0, ns)` : constraint matrix for state variables
 - `F  = zeros(0, nu)` : constraint matrix for input variables
+- `K  = nothing`      : constraint matrix for state elimination
```
sshin23: constraint matrix doesn't sound right. How did Kerrigan call it? If they don't have proper name in the paper, I'd call it feedback gain matrix

src/DynamicNLPModels.jl
```diff
@@ -503,126 +771,217 @@ end
...
-    h0  = zeros(1,1)
-    LinearAlgebra.mul!(QAs, block_Q, transpose(s0TAT))
-    LinearAlgebra.mul!(h0, s0TAT, QAs)
-  
-    c0 = h0[1,1] / 2
-  
-    if S != nothing
-        if !iszero(S)
-            block_S = zeros(ns * (N + 1), nu * N)
+    LinearAlgebra.mul!(STB, transpose(block_S), block_B)
```
sshin23: let's use `'` instead of transpose

sshin23 (approved):

sshin23:
Looks good @dlcole3

src/DynamicNLPModels.jl
dlcole3: They don't explicitly name it, but reference it at one point as being a gain matrix

PR #24, opened 2022-06-21
Added `get_u` and `get_s` functions

dlcole3:
Added functions to query the solution. Functions take an input of the solver reference and the model. They return $s$ and $u$ solution values for all sparse cases. They return $u$ solution only for the condensed case with no `K` matrix. Additional functionalities to query the $s$ and $u$ variables in the condensed case will be addressed later.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -230,7 +230,7 @@ If `condense=false`, data is converted to the form
 Resulting `H` and `J` matrices are stored as `QuadraticModels.QPData` within the `LQDynamicModel` struct and 
 variable and constraint limits are stored within `NLPModels.NLPModelMeta`
 
-If `K` is defined, then `u` variables are replaced by `v` variables, and `u` can be queried by functions to be built within `DynamicNLPModels.jl`
+If `K` is defined, then `u` variables are replaced by `v` variables, and `u` can be queried by `get_u` and `get_s` within `DynamicNLPModels.jl`
 
 ---
 
@@ -245,7 +245,7 @@ If `condense=true`, data is converted to the form
 Resulting `H`, `J`, `h`, and `h0` matrices are stored within `QuadraticModels.QPData` as `H`, `A`, `c`, and `c0` attributes respectively
 
 If `K` is defined, then `u` variables are replaced by `v` variables. The bounds on `u` are transformed into algebraic constraints,
-and `u` can be queried by functions to be built within `DynamicNLPModels.jl`
+and `u` can be queried by `get_u` and `get_s` within `DynamicNLPModels.jl`
 """
 function LQDynamicModel(dnlp::LQDynamicData{T,V,M}; condense = false) where {T, V <: AbstractVector{T}, M  <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix{T}}}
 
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -168,6 +168,13 @@ solution_ref_condense_from_data = madnlp(lq_condense_from_data, max_iter=100)
 @test solution_ref_sparse_from_data.solution[(ns * (N + 1) + 1):(ns * (N + 1) + nu*N)] ≈ solution_ref_condense_from_data.solution atol =  1e-6
 
 
+# Test get_u and get_s functions with no K matrix
+s_values = value.(all_variables(model)[1:(ns * (N + 1))])
+u_values = value.(all_variables(model)[(1 + ns * (N + 1)):(ns * (N + 1) + nu * N)])
+
+@test s_values ≈ get_s(solution_ref_sparse, lq_sparse) atol = 1e-7
```
frapac, line 176: Test should be exact there

src/DynamicNLPModels.jl
```diff
@@ -993,6 +993,74 @@ function _build_condensed_G_blocks(block_A, block_B, block_E, block_F, block_K,
...
+            v = solution[(ns * (N + 1) + start_v):(ns * (N + 1) + end_v)]
+
+            LinearAlgebra.mul!(Ks, K, s)
+            LinearAlgebra.axpy!(1, v, Ks)
+
+            u[start_v:end_v] = Ks
+        end
+
+        return u
+    end
```
sshin23, line 1034: what if it's dense?

src/DynamicNLPModels.jl
```diff
@@ -993,6 +993,74 @@ function _build_condensed_G_blocks(block_A, block_B, block_E, block_F, block_K,
...
+    ) where {T, V <: AbstractVector{T}, M1 <: AbstractMatrix{T}, M2 <: AbstractMatrix{T}, M3 <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix}}
+
+    if lqdm.condense == false
+        solution = solver_status.solution
+        ns       = lqdm.dynamic_data.ns
+        N        = lqdm.dynamic_data.N
+
+        s = solution[1:(ns * (N + 1))]
+        return s
+    end
```
sshin23, line 1074: what if it's dense?

sshin23 (changes requested):

src/DynamicNLPModels.jl
```diff
@@ -993,6 +993,74 @@ function _build_condensed_G_blocks(block_A, block_B, block_E, block_F, block_K,
     return (J = G, lcon = vec(block_gl), ucon = vec(block_gu), As0 = As0)
 end
 
+function get_u(
+    solver_status, 
+    lqdm::LQDynamicModel{T, V, M1, M2, M3, MK}
+    ) where {T, V <: AbstractVector{T}, M1 <: AbstractMatrix{T}, M2 <: AbstractMatrix{T}, M3 <: AbstractMatrix{T}, MK <: AbstractMatrix{T}}
+
+    if lqdm.condense == false
```
frapac: what happens if `lqdm.condense == true`?

src/DynamicNLPModels.jl
```diff
@@ -993,6 +993,74 @@ function _build_condensed_G_blocks(block_A, block_B, block_E, block_F, block_K,
...
+        return u
+    end
+end
+
+function get_u(
+    solver_status, 
+    lqdm::LQDynamicModel{T, V, M1, M2, M3, MK}
+    ) where {T, V <: AbstractVector{T}, M1 <: AbstractMatrix{T}, M2 <: AbstractMatrix{T}, M3 <: AbstractMatrix{T}, MK <: Nothing}
+
+    if lqdm.condense == false
```
frapac: This is an anti-pattern (see PEP8 for instance). Prefer:

```julia
if !lqdm.condense
...
```

src/DynamicNLPModels.jl
```diff
@@ -993,6 +993,74 @@ function _build_condensed_G_blocks(block_A, block_B, block_E, block_F, block_K,
...
+        return solver_stats.solution
+    end
+end
+
+function get_s(
+    solver_status, 
+    lqdm::LQDynamicModel{T, V, M1, M2, M3, MK}
+    ) where {T, V <: AbstractVector{T}, M1 <: AbstractMatrix{T}, M2 <: AbstractMatrix{T}, M3 <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix}}
+
+    if lqdm.condense == false
```
frapac: What happens if `lqdm.condense == true`?

frapac (commented):
LGTM. I just have minor comments. 

I am wondering if it would make sense to define two structures, depending on whether we reduce the model or not. 

Like
```julia
LQDynamicModel
CondensedLQDynamicModel
```
and remove the `condense` attribute

src/DynamicNLPModels.jl
dlcole3: If it's dense, we need the `block_K` and `block_B` matrices. I misunderstood you I think. I will add these, but will just have the computation occur a second time, rather than storing the data like we discussed on Slack.

sshin23:
@frapac agreed that it would be good to define separate data structure for dense/sparse case. In terms of the naming, I'd call them

```julia
SparseLQDynamicModel <: AbstractLQDynamicModel
DenseLQDynamicModel <: AbstractLQDynamicModel

```

frapac:
@sshin23 
I agree with the naming you just proposed:
```julia
SparseLQDynamicModel <: AbstractLQDynamicModel
DenseLQDynamicModel <: AbstractLQDynamicModel
```
Just: which one is supposed to implement the condensed model?

sshin23:
@frapac `DenseLQDynamicModel` implements the dense form (states are eliminated), and condensing (eliminating the inequality block) is done internally in MadNLP

src/DynamicNLPModels.jl
sshin23: If we're not supporting it now (which is totally fine at this stage), then it should throw an error and say it is not supported for dense problems

dlcole3:
@sshin23, do you want me to go ahead and make the change to 
```
SparseLQDynamicModel <: AbstractLQDynamicModel
DenseLQDynamicModel <: AbstractLQDynamicModel
```
now, or should I do this later? Over Slack, it sounded like you suggested doing this later, but I can address it now if desired

test/runtests.jl
dlcole3: @frapac I think this is still inexact because I am comparing the `JuMP` solution to the `LQDynamicModel` solution

dlcole3:
Also, I have been mixed up on `condensed` and `dense`. I think I will rename some of my functions to be `dense` to reflect the proper terminology

sshin23:
@dlcole3 You can work on the new data structure in a separate PR

src/DynamicNLPModels.jl
dlcole3: @sshin23 I updated the code yesterday so that it is now supported. It computes the `A`, `B`, and `K` blocks right now and returns the `s` and `u` values. When I update the data structures, I will save the block matrices within the `DenseLQDynamicModel` data type and can update `get_u` and `get_s` accordingly. Would you prefer I not recalculate these and just have it throw an error for now? I have pushed the support for the dense formulation, so the branch is updated.

test/runtests.jl
sshin23: If both formulations are sparse, it could be exact. @dlcole3 could you check if it's exact?

src/DynamicNLPModels.jl
sshin23: I think computing the whole condensed block again is not the correct approach. I suggest we throw an error if it's dense. We can implement this once we have a separate data structure for `DenseLQDynamicModel` so that we have reference to the reusable condensed block

sshin23 (changes requested):

dlcole3:
It is not exact. It is very close (<1e-10).

test/runtests.jl
dlcole3: It is not exact, but it is close (< 1e-10)

sshin23 (approved):

PR #27, opened 2022-06-22
Added support for non-default matrix types including `CuArray`s 

### LINKED ISSUE #16: Add non-default vector/matrices to the test
Add the following to the test and check whether the model is created correctly.

e.g.,
```julia
s0 = randn(Float32,2)
A = randn(Float32,2,2)
B = randn(Float32,2,2)
Q = randn(Float32,2,2)
R = randn(Float32,2,2)

@test LQDynamicModel(s0,A,B,Q,R,10) isa LQDynamicModel{Float32, Vector{Float32}, SparseArrays.SparseMatrixCSC{Float32, Int64}, SparseArrays.SparseMatrixCSC{Float32, Int64}, Matrix{Float32}}
```

and do the same for `CuArrays`. Currently we can't solve it with MadNLP yet, but later we can also test solving it with MadNLP once MadNLP supports single different precisions

dlcole3:
Updated src code to handle non-default matrix types, mainly by removing `zeros` and using `similar` instead. Updated `runtests.jl` to test on `Float32` types and `CuArray` types. 

These changes also enable forming the dense formulation as `CuArray`s, so that the `H`, `J`, `lcon`, `ucon`, `lvar`, and `uvar` are returned as `CuArray`s when the original arrays passed to `LQDynamicModel` are `CuArray`s.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -6,7 +6,7 @@ import LinearAlgebra
 import SparseArrays
 import SparseArrays: SparseMatrixCSC
 
-export LQDynamicData, LQDynamicModel, get_u, get_s, SparseLQDynamicModel, DenseLQDynamicModel
+export LQDynamicData, SparseLQDynamicModel, DenseLQDynamicModel, get_u, get_s
 
 abstract type AbstractLQDynData{T,V} end
 """
@@ -253,10 +253,9 @@ mutable struct DenseLQDynamicModel{T, V, M1, M2, M3, M4, MK} <:  AbstractDynamic
 end
 
 """
-    LQDynamicModel(dnlp::LQDynamicData; dense=false)      -> SparseLQDynamicModel/DenseLQDynamicModel
-    LQDynamicModel(s0, A, B, Q, R, N; dense = false, ...) -> SparseLQDynamicModel/DenseLQDynamicModel
-A constructor for building a `SparseLQDynamicModel <: QuadraticModels.AbstractQuadraticModel` (if dense = false)
-or a `DenseLQDynamicModel <: QuadraticModels.AbstractQuadraticModel` (if dense = true) from `LQDynamicData`
+    SparseLQDynamicModel(dnlp::LQDynamicData)    -> SparseLQDynamicModel
+    SparseLQDynamicModel(s0, A, B, Q, R, N; ...) -> SparseLQDynamicModel
+A constructor for building a `SparseLQDynamicModel <: QuadraticModels.AbstractQuadraticModel`
 Input data is for the problem of the form 
 ```math
     minimize    \\frac{1}{2} \\sum_{i = 0}^{N-1}(s_i^T Q s_i + 2 u_i^T S^T x_i + u_i^T R u_i) + \\frac{1}{2} s_N^T Qf s_N
@@ -269,7 +268,7 @@ Input data is for the problem of the form
 ```
 ---
 
-If `dense=false`, data is converted to the form 
+Data is converted to the form 
 
 ```math
     minimize    \\frac{1}{2} z^T H z 
@@ -280,33 +279,12 @@ Resulting `H` and `J` matrices are stored as `QuadraticModels.QPData` within the
 variable and constraint limits are stored within `NLPModels.NLPModelMeta`
 
 If `K` is defined, then `u` variables are replaced by `v` variables, and `u` can be queried by `get_u` and `get_s` within `DynamicNLPModels.jl`
-
----
-
-If `dense=true`, data is converted to the form 
-
-```math
-    minimize    \\frac{1}{2} u^T H u + h^T u + h0 
-    subject to  Jz \\le g
-                ul \\le u \\le uu
-```
-
-Resulting `H`, `J`, `h`, and `h0` matrices are stored within `QuadraticModels.QPData` as `H`, `A`, `c`, and `c0` attributes respectively
-
-If `K` is defined, then `u` variables are replaced by `v` variables. The bounds on `u` are transformed into algebraic constraints,
-and `u` can be queried by `get_u` and `get_s` within `DynamicNLPModels.jl`
 """
-function LQDynamicModel(dnlp::LQDynamicData{T,V,M}; dense = false) where {T, V <: AbstractVector{T}, M  <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix{T}}}
-
-    if !dense
+function SparseLQDynamicModel(dnlp::LQDynamicData{T,V,M}) where {T, V <: AbstractVector{T}, M  <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix{T}}}
         _build_sparse_lq_dynamic_model(dnlp)
-    else
-        _build_dense_lq_dynamic_model(dnlp)
-    end
-
 end
 
-function LQDynamicModel(
+function SparseLQDynamicModel(
     s0::V,
     A::M,
     B::M,
@@ -332,8 +310,70 @@ function LQDynamicModel(
         Qf = Qf, S = S, E = E, F = F, K = K, 
         sl = sl, su = su, ul = ul, uu = uu, gl = gl, gu = gu)
     
-    LQDynamicModel(dnlp; dense=dense)
+    SparseLQDynamicModel(dnlp)
+end
 
+"""
+    DenseLQDynamicModel(dnlp::LQDynamicData)    -> DenseLQDynamicModel
+    DenseLQDynamicModel(s0, A, B, Q, R, N; ...) -> DenseLQDynamicModel
+A constructor for building a `DenseLQDynamicModel <: QuadraticModels.AbstractQuadraticModel`
+
+Input data is for the problem of the form 
+```math
+    minimize    \\frac{1}{2} \\sum_{i = 0}^{N-1}(s_i^T Q s_i + 2 u_i^T S^T x_i + u_i^T R u_i) + \\frac{1}{2} s_N^T Qf s_N
+    subject to  s_{i+1} = A s_i + B u_i  for i=0, 1, ..., N-1
+                u_i = Kx_i + v_i  \\forall i = 0, 1, ..., N - 1
+                gl \\le E s_i + F u_i \\le gu for i = 0, 1, ..., N-1            
+                sl \\le s \\le su
+                ul \\le u \\le uu
+                s_0 = s0
+```
+---
+
+Data is converted to the form 
+
+```math
+    minimize    \\frac{1}{2} u^T H u + h^T u + h0 
+    subject to  Jz \\le g
+                ul \\le u \\le uu
+```
+
+Resulting `H`, `J`, `h`, and `h0` matrices are stored within `QuadraticModels.QPData` as `H`, `A`, `c`, and `c0` attributes respectively
+
+If `K` is defined, then `u` variables are replaced by `v` variables. The bounds on `u` are transformed into algebraic constraints,
+and `u` can be queried by `get_u` and `get_s` within `DynamicNLPModels.jl`
+"""
+function DenseLQDynamicModel(dnlp::LQDynamicData{T,V,M}) where {T, V <: AbstractVector{T}, M  <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix{T}}}
+    _build_dense_lq_dynamic_model(dnlp)
+end
+
+function DenseLQDynamicModel(
+    s0::V,
+    A::M,
+    B::M,
+    Q::M,
+    R::M,
+    N;
+    Qf::M = Q, 
+    S::M  = (similar(Q, size(Q, 1), size(R, 1)) .= 0),
+    E::M  = (similar(Q, 0, length(s0)) .= 0),
+    F::M  = (similar(Q, 0, size(R, 1)) .= 0),
+    K::MK = nothing,
+    sl::V = (similar(s0) .= -Inf),
+    su::V = (similar(s0) .=  Inf),
+    ul::V = (similar(s0, size(R, 1)) .= -Inf),
+    uu::V = (similar(s0, size(R, 1)) .=  Inf),
+    gl::V = (similar(s0, size(E, 1)) .= -Inf), 
+    gu::V = (similar(s0, size(F, 1)) .= Inf), 
+    dense=false
+) where {T, V <: AbstractVector{T}, M <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix{T}}}
+
+    dnlp = LQDynamicData(
+        s0, A, B, Q, R, N; 
+        Qf = Qf, S = S, E = E, F = F, K = K, 
+        sl = sl, su = su, ul = ul, uu = uu, gl = gl, gu = gu)
+    
+    DenseLQDynamicModel(dnlp)
 end
 
 function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where {T, V <: AbstractVector{T}, M  <: AbstractMatrix{T}, MK <: Nothing}
@@ -399,8 +439,6 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where 
         uvar[((N + 1) * ns + (j - 1) * nu + 1):((N + 1) * ns + j * nu)] .= uu
     end
     
-
-
     SparseLQDynamicModel(
         NLPModels.NLPModelMeta(
         nvar,
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -494,8 +494,8 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where 
     lvar[1:ns] .= s0
     uvar[1:ns] .= s0
 
-    ucon  = (similar(Q, ns * N + N * length(gl)) .= 0)
-    lcon  = (similar(Q, ns * N + N * length(gl)) .= 0)
+    ucon  = (similar(Q, ns * N + N * length(gl) + length(lcon3)) .= 0)
+    lcon  = (similar(Q, ns * N + N * length(gl) + length(lcon3)) .= 0)
 
     ncon  = size(J, 1)
     nnzj = length(J.rowval)
@@ -509,8 +509,12 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where 
         ucon[(ns * N + 1 + (i -1) * length(gl)):(ns * N + i * length(gl))] .= gu
     end
 
-    lcon = vcat(lcon, lcon3)
-    ucon = vcat(ucon, ucon3)
+    if length(lcon3) > 0
+        lcon[(1 + ns * N + N * length(gl)):end] .= lcon3
+        ucon[(1 + ns * N + N * length(gl)):end] .= ucon3
+    end
+
+
 
     c0 = eltype(Q)(0.0)
     c  = (similar(Q, nvar) .= 0)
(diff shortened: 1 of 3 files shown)


dlcole3:
I did not realize that having a test on `CUDA.jl` would cause the tests to error out (though this makes a lot of sense). While these are removed form the test, running `LQDynamicModel` on my own machine with `CuArray`s did work and resulted in `H` and `J` being returned as `CuArray`s.

sshin23:
yes, we will need to set up a self-hosted runner. I'll set it up later today

sshin23:
you can set up `runtests.jl` in a way that it runs CUDA test only if one can detect an nvidia GPU. You can use `CUDA.has_cuda_gpu()` to check it

src/DynamicNLPModels.jl
```diff
@@ -603,24 +609,25 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
         end
 
 
         sl = sl[bool_vec]
         su = su[bool_vec]
     end
 
-    lcon2 = (similar(Q, (size(J2, 1), 1)) .= 0)
-    ucon2 = (similar(Q, (size(J2, 1), 1)) .= 0)
+    lcon2 = (similar(Q, size(J2, 1), 1) .= 0)
```
frapac: what happens if `num_real_bounds != length(sl)` and `J2` has not been instantiated on line 612?

src/DynamicNLPModels.jl
```diff
@@ -631,16 +638,30 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
     LinearAlgebra.axpy!(-1, As0_bounds, ucon2)
     LinearAlgebra.axpy!(-1, As0_bounds, lcon2)
 
-    lcon = vcat(lcon, vec(lcon2))
-    ucon = vcat(ucon, vec(ucon2))
+    lcon = (similar(Q, length(lcon1) + length(lcon2)) .= 0)
```
frapac: I would prefer the syntax:
```julia
lcon = similar(Q, n) ; fill!(lcon, 0)
```

src/DynamicNLPModels.jl
```diff
@@ -707,30 +728,30 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+    ucon1 = G_blocks.ucon
+    As0   = G_blocks.As0
 
     # Convert state variable constraints to algebraic constraints
     bool_vec_s        = (su .!= Inf .|| sl .!= -Inf)
     num_real_bounds   = sum(bool_vec_s)
 
     if num_real_bounds == length(sl)
-        J2         = block_B[(1 + ns):end,:]
+        J2         = (similar(Q, ns * N, nu * N) .= 0)
```
frapac: same here, we should not instantiate `J2` in the `if` statement

src/DynamicNLPModels.jl
```diff
@@ -1033,7 +1076,7 @@ function _build_H_blocks(block_Q, block_R, block_A, block_B, block_S, block_K, s
     LinearAlgebra.mul!(hS_term, As0', SKAs0)
     LinearAlgebra.mul!(hQ_term, As0', QAs0)
     
-    h0 = 1 / 2 * hR_term + 1 / 2 * hQ_term + hS_term
+    h0 = hR_term/eltype(block_A)(2) + hQ_term/eltype(block_A)(2) + hS_term
```
frapac: Oulalah! This line is overly complicated in my opinion. I would prefer:
```julia
T = eltype(block_A)
h0 = hR_term ./ T(2) .+ hQ_term ./ T(2) .+ hS_term
```

test/runtests.jl
```diff
@@ -435,13 +435,98 @@ set_gl!(lq_dense, 3, rand_val)
 @test get_gl(lq_dense) == gltest
 
 
-# Test non-default vector/matrix types
+# Test non-default vector/matrix on GenericArrays
```
frapac, line 438: Nice!

test/runtests.jl
```diff
@@ -435,13 +435,98 @@ set_gl!(lq_dense, 3, rand_val)
...
+E = Test.GenericArray(E)
+F = Test.GenericArray(F)
+gl = Test.GenericArray(gl)
+gu = Test.GenericArray(gu)
+sl = Test.GenericArray(sl)
+su = Test.GenericArray(su)
+ul = Test.GenericArray(ul)
+uu = Test.GenericArray(uu)
+
+@test LQDynamicModel(s0, A, B, Q, R, 10; S = S, E = E, F = F, gl = gl, gu = gu, ul = ul, uu = uu, sl = sl, su = su,  dense=true) isa DenseLQDynamicModel{Float32, GenericArray{Float32, 1}, GenericArray{Float32, 2}, GenericArray{Float32, 2}, GenericArray{Float32, 2}, SparseMatrixCSC{Float32, Int64}, Nothing}
```
frapac: Ouch, can we split this line?

test/runtests.jl
```diff
@@ -435,13 +435,98 @@ set_gl!(lq_dense, 3, rand_val)
...
+@test LQDynamicModel(s0, A, B, Q, R, 10; K = K, S = S, E = E, F = F, gl = gl, gu = gu, ul = ul, uu = uu, sl = sl, su = su,  dense=true) isa DenseLQDynamicModel{Float32, GenericArray{Float32, 1}, GenericArray{Float32, 2}, GenericArray{Float32, 2}, GenericArray{Float32, 2}, SparseMatrixCSC{Float32, Int64}, GenericArray{Float32, 2}}
+@test LQDynamicModel(s0, A, B, Q, R, 10; S = S, E = E, F = F, gl = gl, gu = gu, ul = ul, uu = uu, sl = sl, su = su,  dense=false) isa SparseLQDynamicModel{Float32, GenericArray{Float32, 1}, SparseMatrixCSC{Float32, Int64}, SparseMatrixCSC{Float32, Int64}, GenericArray{Float32, 2}, Nothing}
+@test LQDynamicModel(s0, A, B, Q, R, 10; K = K, S = S, E = E, F = F, gl = gl, gu = gu, ul = ul, uu = uu, sl = sl, su = su,  dense=false) isa SparseLQDynamicModel{Float32, GenericArray{Float32, 1}, SparseMatrixCSC{Float32, Int64}, SparseMatrixCSC{Float32, Int64}, GenericArray{Float32, 2}, GenericArray{Float32, 2}}
+
+
+# Test with CUDA Array type
+if CUDA.has_cuda_gpu()
```
frapac, line 482: Nice!

frapac (changes requested):
It looks good to me, I just noticed one or two minor issues!

src/DynamicNLPModels.jl
dlcole3: I think it would still be instantied in the `else` statement, but it is redundant. I agree that I should put it outside the `if` statements.

test/runtests.jl
dlcole3: Yeah, this line was pretty ugly... It is split now :)

src/DynamicNLPModels.jl
dlcole3: @frapac, if I am just initiating an array to fill it through `mul!`, I do not need to do `fill!()`, correct?

src/DynamicNLPModels.jl
```diff
@@ -6,7 +6,7 @@ import LinearAlgebra
 import SparseArrays
 import SparseArrays: SparseMatrixCSC
 
-export LQDynamicData, LQDynamicModel, get_u, get_s
+export LQDynamicData, LQDynamicModel, get_u, get_s, SparseLQDynamicModel, DenseLQDynamicModel
```
sshin23: How about we entirely get rid of `LQDynamicModel`?

src/DynamicNLPModels.jl
```diff
@@ -361,22 +364,24 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
     H   = _build_H(Q, R, N; Qf = Qf, S = S)
     J1  = _build_sparse_J1(A, B, N)
     J2  = _build_sparse_J2(E, F, N)
-    J   = vcat(J1, J2)
-    c0 = 0.0
+
+    J = vcat(J1, J2)
+
+    c0  = eltype(Q)(0.0)
```
sshin23: `zero(eltype(Q))`

src/DynamicNLPModels.jl
```diff
@@ -361,22 +364,24 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
...
 
     
     nvar = ns * (N + 1) + nu * N
-    c  = zeros(nvar)
+    c  = similar(Q, nvar); fill!(c, 0)
     
-    lvar  = zeros(nvar)
-    uvar  = zeros(nvar)
+    lvar  = similar(Q, nvar); fill!(lvar, 0)
+    uvar  = similar(Q, nvar); fill!(uvar, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., `s0`

src/DynamicNLPModels.jl
```diff
@@ -361,22 +364,24 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
...
 
     lvar[1:ns] .= s0
     uvar[1:ns] .= s0
 
 
-    ucon  = zeros(ns * N + N * length(gl))
-    lcon  = zeros(ns * N + N * length(gl))
+    lcon  = similar(Q, ns * N + N * length(gl)); fill!(lcon, 0)
+    ucon  = similar(Q, ns * N + N * length(gl)); fill!(ucon, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -400,6 +405,7 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
     SparseLQDynamicModel(
         NLPModels.NLPModelMeta(
         nvar,
+        x0   = (similar(Q, nvar) .= 0),#zeros(eltype(Q), nvar),
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -477,19 +488,20 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
...
-    uvar  = fill(Inf, nvar)
+    lvar  = similar(Q, nvar); fill!(lvar, -Inf)
+    uvar  = similar(Q, nvar); fill!(uvar, Inf)
 
     lvar[1:ns] .= s0
     uvar[1:ns] .= s0
 
-    ucon  = zeros(ns * N + N * length(gl))
-    lcon  = zeros(ns * N + N * length(gl))
+    lcon  = similar(Q, ns * N + N * length(gl) + length(lcon3)); fill!(lcon, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -477,19 +488,20 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
...
+    J = vcat(J1, J2)
+    J = vcat(J, J3)
+
 
     nvar = ns * (N + 1) + nu * N
     
-    lvar  = fill(-Inf, nvar)
-    uvar  = fill(Inf, nvar)
+    lvar  = similar(Q, nvar); fill!(lvar, -Inf)
+    uvar  = similar(Q, nvar); fill!(uvar, Inf)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -503,15 +515,20 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
...
+        lcon[(1 + ns * N + N * length(gl)):end] .= lcon3
+        ucon[(1 + ns * N + N * length(gl)):end] .= ucon3
+    end
+
+
 
-    c0 = 0.0
-    c  = zeros(nvar)
+    c0 = eltype(Q)(0.0)
+    c  = similar(Q, nvar); fill!(c, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -503,15 +515,20 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
...
+    if length(lcon3) > 0
+        lcon[(1 + ns * N + N * length(gl)):end] .= lcon3
+        ucon[(1 + ns * N + N * length(gl)):end] .= ucon3
+    end
+
+
 
-    c0 = 0.0
-    c  = zeros(nvar)
+    c0 = eltype(Q)(0.0)
```
sshin23: `zero(eltype(Q))`

src/DynamicNLPModels.jl
```diff
@@ -503,15 +515,20 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
...
+    c  = similar(Q, nvar); fill!(c, 0)
 
     SparseLQDynamicModel(
         NLPModels.NLPModelMeta(
         nvar,
+        x0   = (similar(Q, nvar) .= 0),
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -573,18 +590,17 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
-    ucon = G_blocks.ucon
-    As0  = G_blocks.As0
+    J1    = G_blocks.J
+    lcon1 = G_blocks.lcon
+    ucon1 = G_blocks.ucon
+    As0   = G_blocks.As0
 
-    lvar = zeros(nu * N)
-    uvar = zeros(nu * N)
+    lvar = similar(Q, nu * N); fill!(lvar, -Inf)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -595,25 +611,25 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
     bool_vec        = (su .!= Inf .|| sl .!= -Inf)
     num_real_bounds = sum(bool_vec)
 
+    J2         = similar(Q, num_real_bounds * N, nu * N); fill!(J2, 0)
+    As0_bounds = similar(Q, num_real_bounds * N, 1); fill!(As0_bounds, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -595,25 +611,25 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
 
-
         sl = sl[bool_vec]
         su = su[bool_vec]
     end
 
-    lcon2 = zeros(size(J2, 1),1)
-    ucon2 = zeros(size(J2, 1),1)
+    lcon2 = similar(Q, size(J2, 1), 1); fill!(lcon2, 0)
+    ucon2 = similar(Q, size(J2, 1), 1); fill!(ucon2, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -699,30 +730,31 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+    lcon1 = G_blocks.lcon
+    ucon1 = G_blocks.ucon
+    As0   = G_blocks.As0
 
     # Convert state variable constraints to algebraic constraints
     bool_vec_s        = (su .!= Inf .|| sl .!= -Inf)
     num_real_bounds   = sum(bool_vec_s)
 
+    J2         = similar(Q, num_real_bounds * N, nu * N); fill!(J2, 0)
+    As0_bounds = similar(Q, num_real_bounds * N, 1); fill!(As0_bounds, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -733,35 +765,41 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
     bool_vec_u       = (ul .!= -Inf .|| uu .!= Inf)
     num_real_bounds = sum(bool_vec_u)
 
-    KBI  = zeros(nu * N, nu * N)
-    KAs0 = zeros(nu * N, 1)
-    I_mat = 1.0Matrix(LinearAlgebra.I, nu * N, nu * N)
+    KBI  = similar(Q, nu * N, nu * N)
+    KAs0 = similar(Q, nu * N, 1)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -733,35 +765,41 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+    for i in 1:(nu * N)
+        I_mat[i, i] = 1.0
+    end
 
     LinearAlgebra.mul!(KBI, block_K, block_B)
     LinearAlgebra.axpy!(1, I_mat, KBI)
     LinearAlgebra.mul!(KAs0, block_K, As0)
 
+    J3          = similar(Q, num_real_bounds * N, nu * N); fill!(J3, 0)
+    KAs0_bounds = similar(Q, num_real_bounds * N, 1); fill!(KAs0_bounds, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

src/DynamicNLPModels.jl
```diff
@@ -1503,8 +1563,8 @@ function _build_sparse_J3(K, N, uu, ul)
 
     full_bool_vec = fill(true, nu * N)
 
-    lcon3 = zeros(nu * N)
-    ucon3 = zeros(nu * N)
+    lcon3 = similar(K, nu * N); fill!(lcon3, 0)
+    ucon3 = similar(K, nu * N); fill!(ucon3, 0)
```
sshin23: here the type should be inferred from the vector given by the user, e.g., s0

sshin23 (changes requested):
Looks good @dlcole3, but one subtle issue:
Remember that our type parameter is `{T,V,M1,...}`, where `V` is type for vector and `M1` is type for matrices. So, when we create a vector, we should infer the type from the given vector, and when we create a matrix, we should infer the type from the given matrix. Currently, this is mixed up in several places. Make sure that the types of created vectors/matrices are derived from the correct source

src/DynamicNLPModels.jl
dlcole3: I think `KAs0_bounds` is actually a matrix in this case. I think some of the LinearAlgebra functions don't work on vectors. I tried to do `mul!(As0, block_A, s0` with `As0 <: AbstractVector{T}`, and I got an error. My understanding is that `mul!()` requires matrices rather than vectors, which is why some of these values are similar to `Q` rather than `s0` since they end up as matrices.

sshin23: could you share the error message?

dlcole3: I tried a couple other things, and it looks like the issue was actually where the `mul!` function returned a 1x3 matrix. In that case, initializing a vector doesn't seem to work, but I was able to work around this one.

sshin23: make sure that `As0` and `s0` are both `AbstractVector`. Note that `Matrix` whose size is either `(3,1)` or `(1,3)` is not a vector

sshin23 (approved):

frapac (approved):

PR #29, opened 2022-06-27
Added example script for dense, sparse, an dense-on-GPU case

### LINKED ISSUE #13: Add example script for using `DenseKKTSystem` in MadNLP
Now that we have dense form LQ problems, we can try to use the full capacity of MadNLP for dense problems. 

Let's create an example directory `./example`, and add a script that solves the dense problem with `DenseKKTSystem` and `MadNLPLapackCPU`.

Check the code here for reference:
https://github.com/MadNLP/MadNLP.jl/blob/master/test/madnlp_dense.jl

dlcole3:
Added a directory `examples/`. Built a function in `build_thinplate.jl` to create a 1D heat transfer problem from the heat transfer equation (a PDE). This function is used for the example script because it is well conditioned and scales well. Added example code in `Madnlp_densekkt_example.jl` to solve the heat transfer equation with the dense solver on both CPU and GPU. Also added code to solve the sparse problem for comparison. Also removed the `CuArray` test in the runtest folder. This seemed redundant with the `GenericArray` test. Further, my understanding is that solving on GPU just requires that the Hessian and Jacobian be `CuArray`s


src/DynamicNLPModels.jl
```diff
@@ -144,20 +144,20 @@ function LQDynamicData(
         error("size of Q is not consistent with length of s0")
     end
 
-    if !(sl  <= su)
+    if sum(sl .>= su) > 0
```
sshin23: Let's implement something like this
```julia
function _cmp_arr(op,A,B)
    for i=1:length(A)
        !op(A[i],B[i]) && return false
    end
    return true
end
```

src/DynamicNLPModels.jl
```diff
@@ -6,7 +6,7 @@ import LinearAlgebra
 import SparseArrays
 import SparseArrays: SparseMatrixCSC
 
-export LQDynamicData, SparseLQDynamicModel, DenseLQDynamicModel, get_u, get_s
+export LQDynamicData, SparseLQDynamicModel, DenseLQDynamicModel, get_u, get_s, Block_Matrices
```
sshin23: The type name should be `BlockMatrices`. Or I'd call it `DenseLQDynamicBlocks`. But is it necessary for this to be exported?

examples/Madnlp_densekkt_example.jl
```diff
@@ -0,0 +1,87 @@
...
+end
+
+function convert_to_CUDA!(lqdm::DenseLQDynamicModel)
+    H = CuArray{Float64}(undef, size(lqdm.data.H))
+    J = CuArray{Float64}(undef, size(lqdm.data.A))
+
+    LinearAlgebra.copyto!(H, lqdm.data.H)
+    LinearAlgebra.copyto!(J, lqdm.data.A)
+
+    lqdm.data.H = H
```
sshin23: shouldn't modify the model, but should recreate a model with concrete type parameters

sshin23 (changes requested):

src/DynamicNLPModels.jl
dlcole3: I originally exported these because I was rebuilding a `DenseLQDynamicModel` that contained `CuArrays`. To do this, I need the `BlockMatrices` structure so that I could pass it to `DenseLQDynamicModel`. But I agree that it should be renamed regardless.

sshin23: You can do it without exporting. To me, `BlockMatrices` should be treated as an internal data structure

examples/Madnlp_densekkt_example.jl
```diff
@@ -0,0 +1,87 @@
+using DynamicNLPModels, NLPModels, Random, LinearAlgebra, MadNLP, QuadraticModels, MadNLPGPU, CUDA, SparseArrays
```
frapac: Can you split this long line?

examples/Madnlp_densekkt_example.jl
frapac: True. In general, I prefer to not use `mutable` structure if we can

examples/Madnlp_densekkt_example.jl
```diff
@@ -0,0 +1,87 @@
...
+    lqdm.data.H = H
+    lqdm.data.A = J
+end
+
+# Define attributes of 1 D heat transfer model
+N  = 10
+ns = 5
+nu = 5
+
+dfunc = (i,k)->100*sin(2*pi*(4*i/N-12*k/ns)) + 400
```
frapac: I advise against using this syntax. Can you define a proper function:
```julia
function dfunc(i, j)
...
end
```

examples/Madnlp_densekkt_example.jl
```diff
@@ -0,0 +1,87 @@
...
+
+d = zeros(ns, N+1)
+
+for i in 1:(N+1)
+    for j in 1:ns
+        d[j,i] = dfunc(i,j)
+    end
+end
```
frapac: Or directly: 
```julia
d = [dfunc(i, j) for i in 1:(N+1), j in 1:ns]
```

examples/Madnlp_densekkt_example.jl
```diff
@@ -0,0 +1,87 @@
...
+
+dx = .1
```
frapac: style guide: prefer writing explicitly `0.1` instead of `.1`. See [here](https://jump.dev/JuMP.jl/dev/developers/style/)

examples/Madnlp_densekkt_example.jl
```diff
@@ -0,0 +1,87 @@
...
+    copyto!(jac, J)
+end
+
+function MadNLP.hess_dense!(nlp::DenseLQDynamicModel{T, V, M1, M2, M3}, x, w1l, hess; obj_weight = 1.0) where {T, V, M1<: Matrix, M2 <: Matrix, M3 <: Matrix}
+    NLPModels.increment!(nlp, :neval_hess)
+    H = nlp.data.H
+    copyto!(hess, H)
+end
+
+function convert_to_CUDA!(lqdm::DenseLQDynamicModel)
```
frapac: Suggestion: rename as `convert_to_cuda!`

examples/build_thinplate.jl
```diff
@@ -0,0 +1,106 @@
+function build_thinplate(ns, nu, N, dx, dt; d = ones(ns, N + 1).*300, Tbar = 300, dense::Bool = true, sl = -Inf, su = Inf)
+    Q = 1.0Matrix(LinearAlgebra.I, ns, ns)
```
frapac: style guide: prefer `Q = 1.0 * Matrix(...)`
The other syntax is the root to confusion

In general, I strongly advise **against** using it.

examples/build_thinplate.jl
```diff
@@ -0,0 +1,106 @@
+function build_thinplate(ns, nu, N, dx, dt; d = ones(ns, N + 1).*300, Tbar = 300, dense::Bool = true, sl = -Inf, su = Inf)
```
frapac: suggestion: `d = fill(300.0, ns, N+1)`
`Tbar = 300.0` (it's a float)

examples/build_thinplate.jl
```diff
@@ -0,0 +1,106 @@
...
+            A[ns, ns - 1] = (dt * conduction_constant)
+        else
+            A[i, i]     = (-2 * dt * conduction_constant + 1)
+            A[i, i - 1] = (dt * conduction_constant)
+            A[i, i + 1] = (dt * conduction_constant)
+
+        end
+    end
+
+    s0 = fill(Float64(Tbar), ns)
```
frapac: see comment above: Tbar should be specified as a float from the beginning

examples/build_thinplate.jl
```diff
@@ -0,0 +1,106 @@
...
+        for i in 1:N
+            block_Q[(1 + (i - 1) * ns):(ns * i), (1 + (i - 1) * ns):(ns * i)] = Q
+        end
+        block_Q[(1 + ns * N):end, (1 + ns * N):end] = Qf
+    end
+    
+    
+    dvec = vec(d)
+    Qd  = similar(dvec)
+    dQd = zeros(1,1)
```
frapac: we assume that `dQd` is a float implicitly, right? why instantiating it as an array?

src/DynamicNLPModels.jl
frapac: See also the function `all`: 
```julia
all(sl .<= su)
```
(but this solution leads to the allocation of one array)

src/DynamicNLPModels.jl
```diff
@@ -696,7 +694,7 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
 
     nvar = nu * N
     nnzj = size(J, 1) * size(J, 2)
-    nnzh = sum(LinearAlgebra.LowerTriangular(H) .!= 0)
+    nnzh = Int(size(H, 1) * (size(H,1) + 1) / 2)
```
frapac: Prefer using `floor(Int, yournumber)`

frapac (changes requested):

frapac (approved):
Look good to me!

sshin23 (approved):

PR #31, opened 2022-07-06
Improved `DenseLQDynamicModel` construction to require less memory and build faster

dlcole3:
Reformulated source code to build the Hessian and Jacobian a block at a time for the dense model. Previously, all block matrices were built, stored, and then multiplied together, but this was very inefficient as many of the multiplications were redundant. Instead, each block within the block matrices were multiplied together only when needed within for loops. 

This PR includes:
- redefining `DenseLQDynamicBlocks` to only save the `A` block and the `B` block. The latter block only contains the first column of the actual block `B` matrix, and it does not contain the first rows of zeros. This also removed any sparse matrices, which were inefficient within `LinearAlgebra.mul!`
- redefining `_build_block_matrices` to match `DenseLQDynamicBlocks`.
- redefining `_build_H_blocks` 
- redefining `_build_G_blocks`. 
- defining a new function, `_build_dense_J1` to form the Jacobian from the `G` blocks
- updating `build_dense_lq_dynamic_model` to be consistent with these changes
- updating `get_s` and `get_u` for working with the new block `B` matrix

This PR does not include updating the sparse formulation (which is also somewhat inefficient) or redefining the internal functions to be non-allocating. These changes are still needed but will be met with other PRs.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -580,10 +580,12 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T, 
 
     nc = size(E, 1)
 
+    bool_vec_s        = (su .!= Inf .|| sl .!= -Inf)
+    num_real_bounds_s = sum(bool_vec_s)
+
     dense_blocks = _build_block_matrices(A, B, K, N)
     block_A  = dense_blocks.A
     block_B  = dense_blocks.B
-
 
     H_blocks = _build_H_blocks(Q, R, block_A, block_B, S,Qf, K, s0, N)
 
@@ -591,13 +593,14 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T, 
     c0 = H_blocks.c0
 
     G  = _init_similar(Q, nc * N, nu, T)
-    J1 = _init_similar(Q, nc * N, nu * N, T)
+    J  = _init_similar(Q, nc * N + num_real_bounds_s * N, nu * N, T)
+    As0_bounds = _init_similar(s0, num_real_bounds_s * N, T)
 
     dl = repeat(gl, N)
     du = repeat(gu, N)
     
     _set_G_blocks!(G, dl, du, block_B, block_A, s0, E, F, K, N)
-    _set_J1_dense!(J1, G, N)
+    _set_J1_dense!(J, G, N)
 
     As0 = _init_similar(s0, ns * (N + 1), T)
     LinearAlgebra.mul!(As0, block_A, s0)
@@ -606,29 +609,24 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T, 
     uvar = repeat(uu, N)
 
     # Convert state variable constraints to algebraic constraints
-    bool_vec        = (su .!= Inf .|| sl .!= -Inf)
-    num_real_bounds = sum(bool_vec)
-
-    J2         = _init_similar(Q, num_real_bounds * N, nu * N, T)
-    As0_bounds = _init_similar(s0, num_real_bounds * N, T)
-
-    if num_real_bounds == length(sl)
+    offset_s = N * nc
+    if num_real_bounds_s == length(sl)
         As0_bounds .= As0[(1 + ns):ns * (N + 1)]
         for i in 1:N
-            J2[(1 + (i - 1) * ns):ns * N, (1 + nu * (i - 1)):(nu * i)] = @view(block_B[1:(ns * (N - i + 1)),:])
+            J[(offset_s + 1 + (i - 1) * ns):(offset_s + ns * N), (1 + nu * (i - 1)):(nu * i)] = @view(block_B[1:(ns * (N - i + 1)),:])
         end
     else        
         for i in 1:N
-            row_range = (1 + (i - 1) * num_real_bounds):(i * num_real_bounds)
-            As0_bounds[row_range] .= As0[(1 + ns * i):(ns * (i + 1))][bool_vec]
+            row_range = (1 + (i - 1) * num_real_bounds_s):(i * num_real_bounds_s)
+            As0_bounds[row_range] .= As0[(1 + ns * i):(ns * (i + 1))][bool_vec_s]
 
             for j in 1:(N - i + 1)
-                J2[(1 + (i + j - 2) * num_real_bounds):((i + j - 1) * num_real_bounds), (1 + nu * (i - 1)):(nu * i)] = @view(block_B[(1 + (j - 1) * ns):(j * ns), :][bool_vec, :])
+                J[(offset_s + 1 + (i + j - 2) * num_real_bounds_s):(offset_s + (i + j - 1) * num_real_bounds_s), (1 + nu * (i - 1)):(nu * i)] = @view(block_B[(1 + (j - 1) * ns):(j * ns), :][bool_vec_s, :])
             end
         end
 
-        sl = sl[bool_vec]
-        su = su[bool_vec]
+        sl = sl[bool_vec_s]
+        su = su[bool_vec_s]
     end
 
     lcon2 = repeat(sl, N)
@@ -644,15 +642,8 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T, 
     ucon[1:length(du)] = du
 
     if length(lcon2) > 0
-        lcon[(1 + length(dl)):(length(dl) + size(J2, 1))] = lcon2
-        ucon[(1 + length(du)):(length(du) + size(J2, 1))] = ucon2
-    end
-
-    J   = _init_similar(Q, size(J1, 1) + size(J2, 1), size(J1, 2), T)
-    J[1:(size(J1, 1)), :] = J1
-
-    if size(J2, 1) > 0
-        J[(1 + size(J1, 1)):(size(J1, 1) + size(J2, 1)), :] = J2
+        lcon[(1 + length(dl)):(length(dl) + num_real_bounds_s * N)] = lcon2
+        ucon[(1 + length(du)):(length(du) + num_real_bounds_s * N)] = ucon2
     end
 
     nvar = nu * N
@@ -725,50 +716,19 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T, 
     H  = H_blocks.H
     c0 = H_blocks.c0
 
-    G   = _init_similar(Q, nc * N, nu, T)
-    J1  = _init_similar(Q, nc * N, nu * N, T)
-    As0 = _init_similar(s0, ns * (N + 1), T)
-
-
-    dl = repeat(gl, N)
-    du = repeat(gu, N)
-
-    
-    _set_G_blocks!(G, dl, du, block_B, block_A, s0, E, F, K, N)
-    _set_J1_dense!(J1, G, N)
-
-    LinearAlgebra.mul!(As0, block_A, s0)
-
-    # Convert state variable constraints to algebraic constraints
     bool_vec_s        = (su .!= Inf .|| sl .!= -Inf)
-    num_real_bounds   = sum(bool_vec_s)
+    num_real_bounds_s   = sum(bool_vec_s)
 
-    J2         = _init_similar(Q, num_real_bounds * N, nu * N, T)
-    As0_bounds = _init_similar(s0, num_real_bounds * N, T)
-
-
-    if num_real_bounds == length(sl)
-        As0_bounds .= As0[(1 + ns):ns * (N + 1)]
-        for i in 1:N
-            J2[(1 + (i - 1) * ns):ns * N, (1 + nu * (i - 1)):(nu * i)] = @view(block_B[1:(ns * (N - i + 1)),:])
-        end
-    else        
-        for i in 1:N
-            row_range = (1 + (i - 1) * num_real_bounds):(i * num_real_bounds)
-            As0_bounds[row_range] = As0[(1 + ns * i):(ns * (i + 1))][bool_vec_s]
-
-            for j in 1:(N - i + 1)
-                J2[(1 + (i + j - 2) * num_real_bounds):((i + j - 1) * num_real_bounds), (1 + nu * (i - 1)):(nu * i)] = @view(block_B[(1 + (j - 1) * ns):(j * ns), :][bool_vec_s, :])
-            end
-        end
-
-        sl = sl[bool_vec_s]
-        su = su[bool_vec_s]
-    end
-
-    # Convert bounds on u to algebraic constraints
     bool_vec_u       = (ul .!= -Inf .|| uu .!= Inf)
-    num_real_bounds  = sum(bool_vec_u)
+    num_real_bounds_u  = sum(bool_vec_u)
+
+
+
+    G   = _init_similar(Q, nc * N, nu, T)
+    J   = _init_similar(Q, (nc + num_real_bounds_s + num_real_bounds_u) * N, nu * N, T)
+    As0 = _init_similar(s0, ns * (N + 1), T)
+    As0_bounds = _init_similar(s0, num_real_bounds_s * N, T)
+    KAs0_bounds = _init_similar(s0, num_real_bounds_u * N, T)
 
     KBI        = _init_similar(Q, nu * N, nu, T)
     KAs0       = _init_similar(s0, nu * N, T)
@@ -778,7 +738,37 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T, 
     I_mat = _init_similar(Q, nu, nu, T)
 
     I_mat[LinearAlgebra.diagind(I_mat)] .= T(1)
-    
+
+    dl = repeat(gl, N)
+    du = repeat(gu, N)
+
+    _set_G_blocks!(G, dl, du, block_B, block_A, s0, E, F, K, N)
+    _set_J1_dense!(J, G, N)
+
+    LinearAlgebra.mul!(As0, block_A, s0)
+
+    # Convert state variable constraints to algebraic constraints
+    offset_s = nc * N
+    if num_real_bounds_s == length(sl)
+        As0_bounds .= As0[(1 + ns):ns * (N + 1)]
+        for i in 1:N
+            J[(offset_s + 1 + (i - 1) * ns):(offset_s + ns * N), (1 + nu * (i - 1)):(nu * i)] = @view(block_B[1:(ns * (N - i + 1)),:])
+        end
+    else        
+        for i in 1:N
+            row_range = (1 + (i - 1) * num_real_bounds_s):(i * num_real_bounds_s)
+            As0_bounds[row_range] = As0[(1 + ns * i):(ns * (i + 1))][bool_vec_s]
+
+            for j in 1:(N - i + 1)
+                J[(offset_s + 1 + (i + j - 2) * num_real_bounds_s):(offset_s + (i + j - 1) * num_real_bounds_s), (1 + nu * (i - 1)):(nu * i)] = @view(block_B[(1 + (j - 1) * ns):(j * ns), :][bool_vec_s, :])
+            end
+        end
+
+        sl = sl[bool_vec_s]
+        su = su[bool_vec_s]
+    end
+
+    # Convert bounds on u to algebraic constraints
     for i in 1:N
         if i == 1
             KB = I_mat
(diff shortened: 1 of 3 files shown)


src/DynamicNLPModels.jl
```diff
@@ -115,9 +117,9 @@ function LQDynamicData(
     N;
 
     Qf::M = Q, 
-    S::M  = (similar(Q, size(Q, 1), size(R, 1)) .= 0),
-    E::M  = (similar(Q, 0, length(s0)) .= 0),
-    F::M  = (similar(Q, 0, size(R, 1)) .= 0),
+    S::M  = (similar(Q, size(Q, 1), size(R, 1)) .= 0.0),
```
frapac: This syntax is too heavy and hard to parse. Prefer: 
```julia
S = similar(Q, n, m) ; fill!(S, 0,0)
```

src/DynamicNLPModels.jl
```diff
@@ -865,23 +876,23 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
-        ucon[(length(ucon1) + length(ucon2) + 1):(length(ucon1) + length(ucon2) + length(ucon3))] .= ucon3
+        lcon[(length(dl) + length(lcon2) + 1):(length(dl) + length(lcon2) + length(lcon3))] .= lcon3
+        ucon[(length(du) + length(ucon2) + 1):(length(du) + length(ucon2) + length(ucon3))] .= ucon3
     end
 
 
     nvar = nu * N
     nnzj = size(J, 1) * size(J, 2)
-    nnzh = sum(LinearAlgebra.LowerTriangular(H) .!= 0)
+    nnzh = floor(Int, size(H, 1) * (size(H,1) + 1) / 2)
```
frapac: For readability, prefer:
```julia
nh = size(H, 1)
nnzh = div(nh * (nh + 1), 2)
```

src/DynamicNLPModels.jl
```diff
@@ -115,9 +117,9 @@ function LQDynamicData(
     N;
 
     Qf::M = Q, 
-    S::M  = (similar(Q, size(Q, 1), size(R, 1)) .= 0),
-    E::M  = (similar(Q, 0, length(s0)) .= 0),
-    F::M  = (similar(Q, 0, size(R, 1)) .= 0),
+    S::M  = (similar(Q, size(Q, 1), size(R, 1)) .= 0.0),
+    E::M  = (similar(Q, 0, length(s0)) .= 0.0),
+    F::M  = (similar(Q, 0, size(R, 1)) .= 0.0),
```
sshin23: zero(T)

src/DynamicNLPModels.jl
```diff
@@ -203,53 +205,30 @@ struct SparseLQDynamicModel{T, V, M1, M2, M3, MK} <:  AbstractDynamicModel{T,V}
...
-    K::M2
-    E::M1
-    F::M1
-    gl::M1
-    gu::M1
+struct DenseLQDynamicBlocks{T, M}
+    A::M
+    B::M
+
 end
```
sshin23: Maybe good to add a docstring of explaining what are these. Or we can even point to the reference

src/DynamicNLPModels.jl
```diff
@@ -203,53 +205,30 @@ struct SparseLQDynamicModel{T, V, M1, M2, M3, MK} <:  AbstractDynamicModel{T,V}
...
 
 function DenseLQDynamicBlocks(
```
sshin23: is this constructor even necessary? can't we just use the default constructor?

src/DynamicNLPModels.jl
```diff
@@ -292,9 +271,9 @@ function SparseLQDynamicModel(
     R::M,
     N;
     Qf::M = Q, 
-    S::M  = (similar(Q, size(Q, 1), size(R, 1)) .= 0),
-    E::M  = (similar(Q, 0, length(s0)) .= 0),
-    F::M  = (similar(Q, 0, size(R, 1)) .= 0),
+    S::M  = (similar(Q, size(Q, 1), size(R, 1)) .= 0.0),
```
sshin23: zero(T)

src/DynamicNLPModels.jl
```diff
@@ -609,30 +588,33 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
 
-    H_blocks = _build_H_blocks(block_Q, block_R, block_A, block_B, block_S, block_K, s0, N, K)
+
+    H_blocks = _build_H_blocks(Q, R, block_A, block_B, S, s0, N, Qf, K)
 
     H  = H_blocks.H
     c0 = H_blocks.c0
 
-    G_blocks = _build_G_blocks(block_A, block_B, block_E, block_F, block_K, block_gl, block_gu, s0, N)
+    G  = similar(Q, nc * N, nu); fill!(G, 0.0)
```
sshin23: `zero(T)`

src/DynamicNLPModels.jl
```diff
@@ -609,30 +588,33 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+    J1 = similar(Q, nc * N, nu * N); fill!(J1, 0.0)
```
sshin23: ok, this is the guy we want to eliminate later

src/DynamicNLPModels.jl
sshin23: `div( ... , 2)`

sshin23 (approved):
Looks good, overall @dlcole3. Have some minor comments

src/DynamicNLPModels.jl
frapac: This is the dense Jacobian, right?

src/DynamicNLPModels.jl
```diff
@@ -609,30 +588,33 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+    dl = similar(s0, nc * N); fill!(dl, 0.0)
+    du = similar(s0, nc * N); fill!(du, 0.0)
+
+    for i in 1:N
+        dl[(1 + nc * (i - 1)): nc * i] = gl
```
frapac: use the broadcast operator here: `dl[...] .= gl`

src/DynamicNLPModels.jl
```diff
@@ -646,17 +628,22 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
-    As0_bounds = similar(s0, num_real_bounds * N); fill!(As0_bounds, 0)
+    J2         = similar(Q, num_real_bounds * N, nu * N); fill!(J2, 0.0)
+    As0_bounds = similar(s0, num_real_bounds * N); fill!(As0_bounds, 0.0)
 
     if num_real_bounds == length(sl)
-        J2        .= block_B[(1 + ns):end,:]
-        As0_bounds .= As0[(1 + ns):end,1]
+        As0_bounds .= As0[(1 + ns):end]
+        for i in 1:N
+            J2[(1 + (i - 1) * ns):end, (1 + nu * (i - 1)):(nu * i)] .= block_B[1:(ns * (N - i + 1)),:]
```
frapac: I would advise replacing `end` by the explicit indexing, to robustify the code

src/DynamicNLPModels.jl
frapac: or more directly: 
```julia
dl = repeat(gl, N)
```

src/DynamicNLPModels.jl
```diff
@@ -749,30 +736,33 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
     gl = dnlp.gl
     gu = dnlp.gu
 
+    nc = size(E, 1)
 
-    dense_blocks = _build_block_matrices(s0, Q, R, A, B, E, F, N, gu, gl, K; Qf = Qf, S = S)
+    dense_blocks = _build_block_matrices(A, B, K, N)
```
frapac, line 704: hehe, just a remark: here `N` is an integer, but uses the same notation in capital as the other matrices

src/DynamicNLPModels.jl
```diff
@@ -749,30 +736,33 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+    J1 = similar(Q, nc * N, nu * N); fill!(J1, 0.0)
+    dl = similar(s0, nc * N); fill!(dl, 0.0)
+    du = similar(s0, nc * N); fill!(du, 0.0)
 
-    J1    = G_blocks.J
-    lcon1 = G_blocks.lcon
-    ucon1 = G_blocks.ucon
-    As0   = G_blocks.As0
+    for i in 1:N
+        dl[(1 + nc * (i - 1)): nc * i] = gl
```
frapac: Use the broadcast operator here, or `repeat`

src/DynamicNLPModels.jl
```diff
@@ -800,30 +795,46 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+        if i == 1
+            KB = I_mat
+        else
+            B_sub_block = block_B[(1 + ns * (i - 2)):ns * (i - 1), :]
+            LinearAlgebra.mul!(KB, K, B_sub_block)
+        end
+        
+        KBI[(1 + nu * (i - 1)):(nu * i),:] = KB
+        LinearAlgebra.mul!(KAs0_temp, K, As0[(1 + ns * (i - 1)):ns * i])
+        KAs0[(1 + nu * (i - 1)):nu * i] = KAs0_temp
```
frapac: broadcast operator

src/DynamicNLPModels.jl
```diff
@@ -800,30 +795,46 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+    KBI  = similar(Q, nu * N, nu)
     KAs0 = similar(s0, nu * N)
+    KAs0_temp = similar(s0, nu)
+    KB   = similar(Q, nu, nu)
 
-    I_mat = similar(Q, nu * N, nu * N); fill!(I_mat, 0)
+    I_mat = similar(Q, nu, nu); fill!(I_mat, 0)
 
-    for i in 1:(nu * N)
+    for i in 1:nu
```
frapac: Or directly: `I_mat[diagind(I_mat)] .= 1.0`

src/DynamicNLPModels.jl
```diff
@@ -800,30 +795,46 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+            J3[(1 + (i - 1) * nu):end, (1 + nu * (i - 1)):(nu * i)] .= KBI[1:(nu * (N - i + 1)),:]
+        end
     else
         for i in 1:N
             row_range   = (1 + (i - 1) * num_real_bounds):(i * num_real_bounds)
-            J3[row_range, :] .= KBI[(1 + nu * (i - 1)):(nu * i), :][bool_vec_u, :]
             KAs0_bounds[row_range]      .= KAs0[(1 + nu * (i - 1)):(nu * i)][bool_vec_u]
+
+            for j in 1:(N - i +1)
+            J3[(1 + (i + j - 2) * num_real_bounds):((i + j - 1) * num_real_bounds), (1 + nu * (i - 1)):(nu * i)] .= KBI[(1 + (j - 1) * nu):(j * nu), :][bool_vec_u, :]
```
frapac: indentation!

src/DynamicNLPModels.jl
```diff
@@ -913,99 +924,55 @@ end
...
     A_klast  = copy(A_k)
     A_knext  = copy(A_k)
-    AB_klast = similar(Q, size(B, 1), size(B, 2)); fill!(AB_klast, 0)
-    AB_k     = similar(Q, size(B, 1), size(B, 2)); fill!(AB_k, 0)
+    AB_klast = similar(A, size(B, 1), size(B, 2)); fill!(AB_klast, 0.0)
+    AB_k     = similar(A, size(B, 1), size(B, 2)); fill!(AB_k, 0.0)
   
     # Fill the A and B matrices
     for i in 1:(N - 1)
         if i == 1
```
frapac: I don't like this syntax. I prefer to initialize the stage `i==1` before entering in the for loop, and then iterates with `for i in 2:(N-1)`

src/DynamicNLPModels.jl
```diff
@@ -913,99 +924,55 @@ end
...
-            block_A[(ns + 1):ns*2, :] = A_k
-            LinearAlgebra.mul!(AB_k, A_k, B)
-            for k in 1:(N-i)
-                row_range = (1 + (k + 1) * ns):((k + 2) * ns)
-                col_range = (1 + (k - 1) * nu):(k * nu)
-                block_B[row_range, col_range] = AB_k
-            end
+            block_A[(ns + 1):ns * 2, :] = A_k
```
frapac: Broadcast!

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
     )
 end
 
-function _build_H_blocks(block_Q, block_R, block_A::M, block_B::M, block_S, block_K, s0, N, K::MK) where {T, M <: AbstractMatrix{T}, MK <: Nothing}
-    As0      = similar(s0, size(block_A, 1))
-    QB       = similar(block_A, size(block_Q, 1), size(block_B, 2))
-    STB      = similar(block_A, size(block_S, 2), size(block_B, 2))
-    B_Q_B    = similar(block_A, size(block_B, 2), size(block_B, 2))
+
+function _build_H_blocks(Q, R, block_A::M, block_B::M, S, s0, N, Qf, K::MK) where {T, M <: AbstractMatrix{T}, MK <: Nothing}
```
frapac: Be careful with the order of the arguments. To be consistent, it's better to put the matrices first, then the scalar arguments.

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
+
+    ns = size(Q, 1)
+    nu = size(R, 1)
+
+    H  = similar(block_A, nu * N, nu * N); fill!(H, 0.0)
+
+    QB  = similar(block_B); fill!(QB, 0.0)
+    QfB = similar(block_B); fill!(QfB, 0.0)
+
+    QAB  = similar(block_A, ns, nu)
```
frapac: When we use `similar`, always initiate the resulting pointer with `0`!

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
+    BQfB = similar(block_B, nu, nu)
+
+    STB = similar(block_B, nu, nu)
+
+    As0 = similar(block_A, ns * (N + 1))
+
+    LinearAlgebra.mul!(As0, block_A, s0)
+
+    for i in 1:N
+        B_sub_block = block_B[(1 + (i - 1) * ns):(i * ns),:]
```
frapac: I guess this lead to unneeded allocations. Please consider using a view here

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
+        LinearAlgebra.mul!(QAB, Q, B_sub_block)
+        LinearAlgebra.mul!(QfAB, Qf, B_sub_block)
+        
+        QB[(1 + (i - 1) * ns):(i * ns), :]  = QAB
```
frapac: Broadcast!

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
+            right_block = block_B[(1 + (j - 1 + i - 1) * ns):((j + i - 1)* ns),:]
+            LinearAlgebra.mul!(BQB, QAB', right_block)
+            LinearAlgebra.mul!(BQfB, QfAB', right_block)
+
+            for k in 1:(N - j - i + 1 + 1)
+                row_range = (1 + nu * (k + (j - 1) - 1)):(nu * (k + (j - 1)))
+                col_range = (1 + nu * (k - 1)):(nu * k)
+
+                if k == N - j - i + 1 + 1
+                    view(H, row_range, col_range) .+= BQfB
```
frapac: genuine question: do we need a view here?

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
+    As0QB = similar(QB, nu)
+    QAs0  = similar(QB, ns)
+    As0S  = similar(block_B, nu)
+
+
+    for i in 1:N
+        fill!(QB_block_vec, 0.0)
+        rows_QB           = 1:(ns * (N - i)) 
+        rows_QfB          = (1 + ns * (N - i)):(ns * (N - i + 1))
+        QB_block_vec[(1 + ns * i):(ns * N), :] = QB[rows_QB, :]
```
frapac: broadcast!

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
+    LinearAlgebra.mul!(QAs0, Qf, As0[(ns * N + 1):end])
+
+    h0 += LinearAlgebra.dot(As0[(ns * N + 1):end], QAs0)
+
+    return (H = H, c = h, c0 = h0 / T(2))
+end
+
+
+
+function _build_H_blocks(Q, R, block_A::M, block_B::M, S::M, s0, N, Qf, K::MK) where {T, M <: AbstractMatrix, MK <: AbstractMatrix{T}}
```
frapac: genuine question: can we use a single `_build_H_blocks` function instead of two? Just to avoid code duplicate.

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
+    # quad term refers to the summation of Q, K^T RK, SK, and K^T S^T that is left and right multiplied by B in the Hessian
+    quad_term = similar(Q, ns, ns); fill!(quad_term, 0.0)
+
+    quad_term_B  = similar(block_B); fill!(quad_term_B, 0.0)
+    QfB = similar(block_B); fill!(QfB, 0.0)
+
+    quad_term_AB = similar(block_A, ns, nu) 
+    QfAB         = similar(block_A, ns, nu)
+
+    RK_STB    = similar(block_B, nu, nu)
```
frapac: also: rule of thumb, for an old-school guy: always allocate the arrays at the beginning of the function (in C, your compiler should tell you to do so)

frapac: if we stick to that convention, it's a lot easier to assess how much memory is allocated inside one function

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
-    return (J = G, lcon = vec(block_gl), ucon = vec(block_gu), As0 = As0)
+
+    EAs0 = similar(block_B, nc)
+    
+    for i in 1:N
+        if i != N
+            B_sub_block = block_B[(1 + (i - 1) * ns):(i * ns),:]
+
+            LinearAlgebra.mul!(EB, E_FK, B_sub_block)
+            G[(1 + nc * i):(nc * (i + 1)), :] = EB
```
frapac, line 1109: broadcast!

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
+end
+
+function _set_G_blocks!(G, dl, du, block_B, block_A, s0, E, F, K::MK, N) where {T, MK <: AbstractMatrix{T}}
+    ns = size(E, 2)
+    nu = size(F, 2)
+    nc = size(E, 1)
 
-    block_E_mul = similar(block_E)
-    LinearAlgebra.copyto!(block_E_mul, block_E)
+    G[1:nc, :] = F
```
frapac, line 1089: broadcast!

src/DynamicNLPModels.jl
```diff
@@ -1015,141 +982,308 @@ function _build_block_matrices(
...
-    LinearAlgebra.axpy!(-1, EAs0, block_gl)
-    LinearAlgebra.axpy!(-1, EAs0, block_gu)
-  
-    return (J = G, lcon = vec(block_gl), ucon = vec(block_gu), As0 = As0)
+
+    EAs0 = similar(block_B, nc)
+    
+    for i in 1:N
+        if i != N
+            B_sub_block = block_B[(1 + (i - 1) * ns):(i * ns),:]
```
frapac: please consider using a view here

frapac (changes requested):
I think this PR is still a work in progress, but most requests are easy to address. 

As a rule of thumb, I would advise
1. Use the broadcast operator consistently and ensure we are not allocating any new array inside a for loop
2. Move all the allocations (`similar`) at the beginning of the functions, so we can quickly assess how much memory is allocated
3. Use the notations from Kerrigan's paper consistently

src/DynamicNLPModels.jl
dlcole3: Not quite. The `G` matrix makes up part of the dense Jacobian. For the sake of memory, I actually just save the first block column of the `G` matrix, and then I fill the Jacobian separately. The Jacobian is also made up of the variable bounds (which become algebraic constraints). I have a couple prepared equations I will show tomorrow.

src/DynamicNLPModels.jl
dlcole3: Woah, I wish I knew about the `repeat` function a long time ago :) I will be adding that in several places. Thank you!

src/DynamicNLPModels.jl
dlcole3: @frapac Yes, I believe so. I played with several things, but the `view` was the one that led to the smallest number of allocations. I honestly don't know why, but when I used other things (like `H[row_range, col_range]`), this was far less efficient and took lots more memory.

src/DynamicNLPModels.jl
dlcole3: I think we can. Previously, `_build_H_blocks` had to have more memory allocation within it to account for more multiplications. However, now the memory allocation difference is negligible under the new formulation. I checked the allocation for a large scale problem, and the two functions with and without `K` used 2.313 GiB vs. 2.311 GiB respectively.

src/DynamicNLPModels.jl
dlcole3: @frapac can I do `fill!(S, 0)` within the function argument? This is setting the default value. I used `.= 0` because I wasn't sure how it would manage a `fill!` function within the function arguments.

dlcole3:
I also updated the `build_thinplate.jl` file in the `examples/` folder. This used the old `DenseLQDynamicBlocks` structure

dlcole3:
@frapac I made one more change after requesting reviews again (sorry, I should've added it before requesting reviews). I realized I am allocating extra memory for the Jacobian. I essentially was buliding the Jacobian twice. I removed the extra `J1`, `J2`, and `J3` matrices in the most recent commit.

src/DynamicNLPModels.jl
```diff
@@ -749,148 +704,148 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
     else
         for i in 1:N
-            row_range   = (1 + (i - 1) * num_real_bounds):(i * num_real_bounds)
-            J3[row_range, :] .= KBI[(1 + nu * (i - 1)):(nu * i), :][bool_vec_u, :]
-            KAs0_bounds[row_range]      .= KAs0[(1 + nu * (i - 1)):(nu * i)][bool_vec_u]
+            row_range              = (1 + (i - 1) * num_real_bounds_u):(i * num_real_bounds_u)
+            KAs0_bounds[row_range] = KAs0[(1 + nu * (i - 1)):(nu * i)][bool_vec_u]
+
+            for j in 1:(N - i +1)
+            J[(offset_u + 1 + (i + j - 2) * num_real_bounds_u):(offset_u + (i + j - 1) * num_real_bounds_u), (1 + nu * (i - 1)):(nu * i)] = @view(KBI[(1 + (j - 1) * nu):(j * nu), :][bool_vec_u, :])
```
frapac: indentation!

frapac (approved):
Just an indentation issue, but otherwise it looks good to merge. 

Good work!

PR #32, opened 2022-07-11
Improved Efficiency of Sparse Model construction

dlcole3:
The sparse model construction was very slow because of several for loops for building the sparse `H` and `J` matrices that contained setting sparse arrays with dense arrays (e.g., `H[..., ...] = Q`). This PR now builds the `H` and `J` matrices by instead building vectors for `colptr`, `rowval`, and `nzval` from the corresponding data, and then using the `SparseMatrixCSC` constructor to create the sparse array from these vectors. The `colptr`, `rowval`, and `nzval` for `H` and `J` are instantiated within the `_build_sparse_lq_dynamic_model` function. The `_set_sparse_H!` and `_set_sparse_J!` functions are then called to set this data. For large problems (N, ns, and nu ~100+), there is significant speedup from these changes.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -379,11 +379,11 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where 
 
     _set_sparse_H!(H_colptr, H_rowval, H_nzval, Q, R, N; Qf = Qf, S = S)
 
-    H = SparseMatrixCSC((N + 1) * ns + nu * N, (N + 1) * ns + nu * N, H_colptr, H_rowval, H_nzval)
+    H = SparseArrays.SparseMatrixCSC((N + 1) * ns + nu * N, (N + 1) * ns + nu * N, H_colptr, H_rowval, H_nzval)
 
     _set_sparse_J!(J_colptr, J_rowval, J_nzval, A, B, E, F, K, N)
 
-    J = SparseMatrixCSC((nc + ns) * N, (N + 1) * ns + nu * N, J_colptr, J_rowval, J_nzval)
+    J = SparseArrays.SparseMatrixCSC((nc + ns) * N, (N + 1) * ns + nu * N, J_colptr, J_rowval, J_nzval)
 
     SparseArrays.dropzeros!(H)
     SparseArrays.dropzeros!(J)
@@ -515,11 +515,11 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where 
     # Get H and J matrices from new matrices
     _set_sparse_H!(H_colptr, H_rowval, H_nzval, new_Q, R, N; Qf = Qf, S = new_S)
 
-    H = SparseMatrixCSC((N + 1) * ns + nu * N, (N + 1) * ns + nu * N, H_colptr, H_rowval, H_nzval)
+    H = SparseArrays.SparseMatrixCSC((N + 1) * ns + nu * N, (N + 1) * ns + nu * N, H_colptr, H_rowval, H_nzval)
 
     _set_sparse_J!(J_colptr, J_rowval, J_nzval, new_A, B, new_E, F, K, bool_vec, N, num_real_bounds)
 
-    J = SparseMatrixCSC(ns * N + nc * N + num_real_bounds * N, (N + 1) * ns + nu * N, J_colptr, J_rowval, J_nzval)
+    J = SparseArrays.SparseMatrixCSC(ns * N + nc * N + num_real_bounds * N, (N + 1) * ns + nu * N, J_colptr, J_rowval, J_nzval)
 
     SparseArrays.dropzeros!(H)
     SparseArrays.dropzeros!(J)
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -366,14 +366,16 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where 
     gl = dnlp.gl
     gu = dnlp.gu
 
-    H   = _build_H(Q, R, N; Qf = Qf, S = S)
-    J1  = _build_sparse_J1(A, B, N)
-    J2  = _build_sparse_J2(E, F, N)
+    nc = size(E, 1)
 
-    J = vcat(J1, J2)
+    H = SparseArrays.sparse([],[],T[], (ns * (N + 1) + nu * N), (ns * (N + 1) + nu * N))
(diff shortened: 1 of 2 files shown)


src/DynamicNLPModels.jl
```diff
@@ -366,14 +367,28 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
     gl = dnlp.gl
     gu = dnlp.gu
 
-    H   = _build_H(Q, R, N; Qf = Qf, S = S)
-    J1  = _build_sparse_J1(A, B, N)
-    J2  = _build_sparse_J2(E, F, N)
+    nc = size(E, 1)
 
-    J = vcat(J1, J2)
+    H_colptr = _init_similar(Int[], ns * (N + 1) + nu * N + 1, Int)
```
frapac: I would prefer we reserve using `_init_similar` for vectors than can be instantiated on the GPU, and use `zeros(Int, m)` if the vector is intended for CPU only.

src/DynamicNLPModels.jl
```diff
@@ -1484,143 +1527,194 @@ function NLPModels.jac_coord!(
...
-        range_Q = (1 + (i - 1) * ns): (i * ns)
-        range_R = (ns * (N + 1) + 1 + (i - 1) * nu):(ns * (N + 1) + i * nu)
-        H[range_Q, range_Q] = Q
-        H[range_R, range_R] = R
-        if S != nothing
-            H[range_Q, range_R] = S
-            H[range_R, range_Q] = S'
+        for j in 1:ns
+            H_nzval[(1 + (i - 1) * (ns^2 + nu * ns) + (j - 1) * (ns + nu)):(ns * j + nu * (j - 1) + (i - 1) * (ns^2 + nu * ns))]  = @view Q[:, j]
+            H_nzval[(1 + (i - 1) * (ns^2 + nu * ns) + j * ns + (j - 1) * nu):((i - 1) * (ns^2 + nu * ns) + j * (ns + nu))] = @view S'[:, j]
```
frapac: Why taking the transpose `S'[:, j]` instead of doing directly `S[j, :]`?

src/DynamicNLPModels.jl
```diff
@@ -1484,143 +1527,194 @@ function NLPModels.jac_coord!(
...
- 1.0  2.0  -1.0    ⋅     ⋅     ⋅   5.0  6.0   ⋅    ⋅    ⋅    ⋅
- 3.0  4.0    ⋅   -1.0    ⋅     ⋅   7.0  8.0   ⋅    ⋅    ⋅    ⋅
-  ⋅    ⋅    1.0   2.0  -1.0    ⋅    ⋅    ⋅   5.0  6.0   ⋅    ⋅
-  ⋅    ⋅    3.0   4.0    ⋅   -1.0   ⋅    ⋅   7.0  8.0   ⋅    ⋅
-```
+    _set_sparse_J!(J_colptr, J_rowval, J_nzval, A, B, E, F, K, bool_vec, N, nb)
+    _set_sparse_J!(J_colptr, J_rowval, J_nzval, A, B, E, F, K, N)
+
+set the data needed to build a SparseArrays.SparseMatrixCSC matrix. J_colptr, J_rowval, and J_nzval
+are set so that they can be passed to SparseMatrixCSC() to obtain the Jacobain, `J`. The Jacobian
```
frapac: Typo: `Jacobain` -> `Jacobian`

src/DynamicNLPModels.jl
```diff
@@ -1484,143 +1527,194 @@ function NLPModels.jac_coord!(
...
+
+If `K` is defined, then this matrix also contains the constraints
+ul <= Kx_i + v_i <= uu
 """
-function _build_sparse_J1(A::M, B::M, N) where M <: AbstractMatrix
+function _set_sparse_J!(
+    J_colptr, J_rowval, J_nzval,
+     A, B, E, F, K::MK, bool_vec,
+     N, nb
+     ) where {T, MK <: AbstractMatrix{T}}
```
frapac: Indentation: the parenthesis `)` should be at position 0

src/DynamicNLPModels.jl
```diff
@@ -1484,143 +1527,194 @@ function NLPModels.jac_coord!(
...
 
-function _build_sparse_J3(K::M, N, uu, ul) where {T, M <: AbstractMatrix{T}}
 
-    # Remove algebraic constraints if u variable is unbounded on both upper and lower ends
-    nu = length(ul)
-    ns = size(K, 2)
+function _set_sparse_J!(
+    J_colptr, J_rowval, J_nzval,
+     A::M, B::M, E, F, K::MK, N
+     ) where {T, M <: AbstractMatrix{T}, MK <: Nothing}
```
frapac: Indentation

frapac (approved):
Nice job! I have only a few minor comments so far.

src/DynamicNLPModels.jl
dlcole3: Haha, I have no good reason. I just didn't think of that, but that makes a lot more sense :) I will adjust that one

src/DynamicNLPModels.jl
sshin23: until we support solving sparse problems on GPU 😬 But at this point no need for using `_init_similar`, since we're not inferring type from the user input

sshin23: Similarly, `_init_similar(T[], ...)` should be replaced by `zeros(T,...)`

src/DynamicNLPModels.jl
```diff
@@ -366,14 +367,28 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where
...
+    _set_sparse_H!(H_colptr, H_rowval, H_nzval, Q, R, N; Qf = Qf, S = S)
+
+    H = SparseArrays.SparseMatrixCSC((N + 1) * ns + nu * N, (N + 1) * ns + nu * N, H_colptr, H_rowval, H_nzval)
 
+    _set_sparse_J!(J_colptr, J_rowval, J_nzval, A, B, E, F, K, N)
+
+    J = SparseArrays.SparseMatrixCSC((nc + ns) * N, (N + 1) * ns + nu * N, J_colptr, J_rowval, J_nzval)
+
+    SparseArrays.dropzeros!(H)
+    SparseArrays.dropzeros!(J)
```
sshin23, line 389: Can we go even further? Can we preallocate the memory based on the sparsity of `Q,R,S`, and remove the `dropzeros!` step?

sshin23: But if this is not obvious & the current implementation is efficient enough, I suggest opening an issue and improving it later

sshin23 (approved):

PR #34, opened 2022-07-12
Added support for Jacobian Linear Operator

dlcole3:
Added a mutable struct `LQJacobianOperator <: AbstractLinearOperator` that contains the data needed for finding $Jx$, $J^Tx$, and $J^T \Sigma J$ from the first `nu` columns of the Jacobian. Added an argument `truncate` to `DenseLQDynamicModel` that calls a different function (`_build_truncated_dense_lq_dynamic_model`). This new function returns the `DenseLQDynamicModel` with the `QPData` containing the `LQJacobianOperator` as the Jacobian rather than a matrix. I also exteded `mul!()` to work with the `LQJacobianOperator` to calculate $Jx$ and $J^Tx$. I also added tests to `runtests.jl` to test this new operator, and I added `LinearOperators.jl` to the dependencies. 

This PR does not include the functions to calculate $J^T \Sigma J$. I will do that on a separate PR. I first wanted to get feedback on the current operator.


src/DynamicNLPModels.jl
```diff
@@ -312,8 +314,12 @@ Resulting `H`, `J`, `h`, and `h0` matrices are stored within `QuadraticModels.QP
 If `K` is defined, then `u` variables are replaced by `v` variables. The bounds on `u` are transformed into algebraic constraints,
 and `u` can be queried by `get_u` and `get_s` within `DynamicNLPModels.jl`
 """
-function DenseLQDynamicModel(dnlp::LQDynamicData{T,V,M}) where {T, V <: AbstractVector{T}, M  <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix{T}}}
-    _build_dense_lq_dynamic_model(dnlp)
+function DenseLQDynamicModel(dnlp::LQDynamicData{T,V,M}; truncated = false) where {T, V <: AbstractVector{T}, M  <: AbstractMatrix{T}, MK <: Union{Nothing, AbstractMatrix{T}}}
```
frapac: I would suggest documenting in the docstring the new optional argument `truncated`

src/DynamicNLPModels.jl
```diff
@@ -333,17 +339,72 @@ function DenseLQDynamicModel(
...
+ - `J1Bx` : vector for storing multiplications when doing `Jx`
+ - `J2Bx` : vector for storing multiplications when doing `Jx`
+ - `J3Bx` : vector for storing multiplications when doing `Jx`
+ - `J1BTx`: vector for storing multiplications when doing `J^T x`
+ - `J2BTx`: vector for storing multiplications when doing `J^T x`
+ - `J3BTx`: vector for storing multiplications when doing `J^T x`
+ - `J1B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
+ - `J2B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
+ - `J3B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
+"""
```
frapac, line 382: Nice docstring!

src/DynamicNLPModels.jl
```diff
@@ -333,17 +339,72 @@ function DenseLQDynamicModel(
...
+mutable struct LQJacobianOperator{T, V, M} <: LinearOperators.AbstractLinearOperator{T}
```
frapac: This structure should not be mutable (as a general rule of thumb, we should avoid using mutable struct wherever possible)

src/DynamicNLPModels.jl
```diff
@@ -333,17 +339,72 @@ function DenseLQDynamicModel(
...
+    Jac::M   # column of Jacobian block matrix
+    N        # number of time steps
```
frapac: Add the type for the following attributes: 
```julia
N::Int
nu::Int
...
```

src/DynamicNLPModels.jl
```diff
@@ -901,6 +962,314 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+        Jac, N, nu, nc, num_real_bounds_s, 0,
+        Jac,
+        _init_similar(s0, nc, T), _init_similar(s0, num_real_bounds_s, T), _init_similar(s0, 0, T),
+        _init_similar(s0, nu, T), _init_similar(s0, nu, T), _init_similar(s0, nu, T),
+        _init_similar(s0, nc, nu, T), _init_similar(s0, num_real_bounds_s, nu, T), _init_similar(s0, 0, nu, T),
+    )
+
+    DenseLQDynamicModel(
+        NLPModels.NLPModelMeta(
+        nvar,
```
frapac: Indentation is missing here (from line 1069 to line 1079)

src/DynamicNLPModels.jl
```diff
@@ -901,6 +962,314 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+        lcon = lcon,
+        ucon = ucon,
+        nnzj = nnzj,
+        nnzh = nnzh,
+        lin = 1:ncon,
+        islp = (ncon == 0);
+        ),
+        NLPModels.Counters(),
+        QuadraticModels.QPData(
+        c0,
```
frapac: same here, indentation is missing

src/DynamicNLPModels.jl
```diff
@@ -901,6 +962,314 @@ function _build_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) where {T,
...
+    ucon3 = repeat(uu, N)
+
+    LinearAlgebra.axpy!(-1, As0_bounds, lcon2)
+    LinearAlgebra.axpy!(-1, As0_bounds, ucon2)
+
+    LinearAlgebra.axpy!(-1, KAs0_bounds, lcon3)
+    LinearAlgebra.axpy!(-1, KAs0_bounds, ucon3)
+
+
+    lcon = _init_similar(s0, size(Jac, 1), T)
```
frapac: I would prefer we allocate everything at the beginning of the function

src/DynamicNLPModels.jl
```diff
@@ -1735,4 +2104,75 @@ function _init_similar(mat, dim1::Number, T=eltype(mat))
     return new_mat
 end
 
+
+function LinearAlgebra.mul!(y::Vector{T},
+    Jac::LQJacobianOperator{T, V, M},
+    x::Vector{T}
+) where {T, V <: AbstractVector{T}, M <: AbstractMatrix{T}}
```
frapac, line 2108: Here, I would enforce `y` and `x` to be of type `V`
```julia
function LinearAlgebra.mul!(y::V,
    Jac::LQJacobianOperator{T, V, M},
    x::V
) where {T, V <: AbstractVector{T}, M <: AbstractMatrix{T}}
```

src/DynamicNLPModels.jl
```diff
@@ -1735,4 +2104,75 @@ function _init_similar(mat, dim1::Number, T=eltype(mat))
...
+end
+
+function LinearAlgebra.mul!(
+    y::Vector{T},
+    Jac::LinearOperators.AdjointLinearOperator{T, LQJacobianOperator{T, V, M}},
+    x::Vector{T}
+) where {T, V <: AbstractVector{T}, M <: AbstractMatrix{T}}
+    fill!(y, zero(T))
+
+    J   = Jac'.Jac
```
frapac: oulala. I understand the syntax is correct, but I think it's a bit too heavy. Can we use a getter to access to the linear operator stored inside `Jac`?

test/runtests.jl
```diff
@@ -67,6 +67,32 @@ solution_ref_dense_from_data  = madnlp(lq_dense_from_data, max_iter=100)
...
+
+Random.seed!(10)
+x = rand(nu * N)
+y = rand(size(lq_dense.data.A, 1))
+x_trunc = copy(x)
+y_trunc = copy(y)
+
+lq_dense_trunc = DenseLQDynamicModel(dnlp;truncated=true)
+
+Jac = lq_dense_trunc.data.A
```
frapac: I would suggest implementing a proper getter to access the Jacobian inside a `DenseLQDynamicModel`

test/runtests.jl
```diff
@@ -67,6 +67,32 @@ solution_ref_dense_from_data  = madnlp(lq_dense_from_data, max_iter=100)
...
+
+LinearAlgebra.mul!(y, lq_dense.data.A, x)
+LinearAlgebra.mul!(y_trunc, Jac, x_trunc)
```
frapac: nice! It's exactly the test I wanted

frapac (changes requested):
Nice! This is exactly what I was looking for. 
So far I only have minor comments. It would be also nice to implement part of the API for the `LQJacobianOperator`. I am thinking of overloading the following functions: 
- `Base.length`
- `Base.size`
- `Base.eltype` 
- `Base.isreal` 
- `LinearOperators.reset!`

src/DynamicNLPModels.jl
dlcole3: @frapac I was concerned about resetting the storage vectors/matrices and whether that would throw an error if this was a `struct` rather than a `mutable struct`. Is changing `J1Bx` inside `mul!` permitted for an immutable `struct`

src/DynamicNLPModels.jl
dlcole3: @frapac What is a "getter"?

frapac: A function `get_operator` for instance, that returns a pointer to the operator `Jac` :)

src/DynamicNLPModels.jl
frapac: Good question. But this is not a concerned for us. Here, `J1Bx` is a pointer, and we don't change the reference once the memory has been allocated. We update only the values in the memory chunk specified by the pointer, not the pointer itself. So in short, `J1Bx` is a mutable structure, but `LQJacobianOperator` does not have to be mutable.

src/DynamicNLPModels.jl
dlcole3: @frapac I changed `nu` and `nc` to be of type `Int`. When I built the `LQDynamicData` struct, I think Sungho suggested leaving `N` without a specific type. I am assuming this is for "user friendliness" so that a user could pass a Float or something else and not run into errors from it. Is there benefit to specifying it to type `Int`?

sshin23: @dlcole3 For functions, we don't need type specifications most of the time. For data structure, it's a different story. We need to specify the type whenever we can to avoid type inference. So here we need to specify it as `Int`

src/DynamicNLPModels.jl
sshin23: `truncated` doesn't sound right to me here. How about `implicit`?

src/DynamicNLPModels.jl
```diff
@@ -1735,4 +2104,75 @@ function _init_similar(mat, dim1::Number, T=eltype(mat))
...
+    J2Bx = Jac.J2Bx
+    J3Bx = Jac.J3Bx
+
+    for i in 1:N
+        sub_B1 = @view J[(1 + (i - 1) * nc):(i * nc), :]
+        sub_B2 = @view J[(1 + nc * N + (i - 1) * nsc):(nc * N + i * nsc), :]
+        sub_B3 = @view J[(1 + nc * N + nsc * N + (i - 1) * nuc):(nc * N + nsc * N + nuc * i), :]
+
+        for j in 1:(N - i + 1)
+            LinearAlgebra.mul!(J1Bx, sub_B1, view(x, (1 + (j - 1) * nu):(j * nu)))
```
sshin23: why do we need `J1Bx`, when you can directly add them to `view(y, (1 + nc * (j + i - 2)):(nc * (j + i - 1) ))` via `mul!(C, A, B, α, β)`?

test/runtests.jl
```diff
@@ -379,6 +542,33 @@ solution_ref_dense_from_data  = madnlp(lq_dense_from_data, max_iter=100)
...
```
sshin23: can we test if `mul!` are non-allocating via `@allocated`? It looks non-allocating right now, but we want to avoid performance degradation in the future

dlcole3: @sshin23 $Jx$ is nonallocating, but $J^T x$ is allocating. I will take a look at that one

dlcole3: @sshin23 @frapac  It looks like the allocations are coming from having to take the Adjoint of the `LQJacobianOperator`. If I do `@allocated Jac_adj = Jac'`, then I get allocation. Is there another way of taking the adjoint that avoids allocation? I cannot seem to find any functions that fit

sshin23: right it will allocate something to make a nested object. You can try doing `Jact = Jac'`, then `@allocated mul!( ..., Jact, ...)`

dlcole3: Ah I see. I was thinking the allocation was coming from taking the Adjoint of the Adjoint, but that is not the case after all. It looks like the $J^Tx$ is not actually allocating then.

sshin23 (approved):

frapac (approved):
LGTM! I would just suggest to rename `get_Jacobian` as `get_jacobian`

PR #35, opened 2022-07-14
Created function `add_jtsj!` and accompanying tests

dlcole3:
Added `add_jtsj!` to the source code to calculate $J^T \Sigma J$ from a `LQJacobianOperator` and a vector, $\Sigma$, and adds this value to a matrix $H$. The code is similar in construct to the `_build_H_blocks` function, but somewhat less efficient. This is because a new block matrix must be created and multiplied every iteration of the for loop, whereas `_build_H_blocks` could reuse the same matrix  several times (e.g., `_build_H_blocks` could calculate $QB$ one time and add it N times to $H$, whereas `add_jtsj!` must calculate separate $\Sigma_i B, \forall i = 1,...,N$).

Also updated `LQJacobianOperator` and removed the `Scaled_Jac` attribute (unnecessary) and changed the `J1B`, `J2B`, and `J3B` matrices to `SJ1`, `SJ2`, and `SJ3` for storing the scaled values of the Jacobian. Also added a function to `runtest.jl` to test `add_jtsj!` and a function to test the `mul` functions with the `LQJacobianOperator`. This latter function was just to simplify the file.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -8,7 +8,7 @@ import LinearOperators
 
 import SparseArrays: SparseMatrixCSC
 
-export LQDynamicData, SparseLQDynamicModel, DenseLQDynamicModel, get_u, get_s, get_jacobian
+export LQDynamicData, SparseLQDynamicModel, DenseLQDynamicModel, get_u, get_s, get_jacobian, add_jtsj!
 
 abstract type AbstractLQDynData{T,V} end
 """
@@ -369,16 +369,10 @@ Attributes
  - `nc` : number of algebraic constraints of the form gl <= Es + Fu <= gu
  - `nsc`: number of bounded state variables
  - `nuc`: number of bounded input variables (if `K` is defined)
- - `scaled_Jac`: placeholder to avoid allocaiton when calculating `Sigma J`
- - `J1Bx` : vector for storing multiplications when doing `Jx`
- - `J2Bx` : vector for storing multiplications when doing `Jx`
- - `J3Bx` : vector for storing multiplications when doing `Jx`
- - `J1BTx`: vector for storing multiplications when doing `J^T x`
- - `J2BTx`: vector for storing multiplications when doing `J^T x`
- - `J3BTx`: vector for storing multiplications when doing `J^T x`
- - `J1B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
- - `J2B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
- - `J3B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
+ - `SJ1`: placeholder for storing data when calculating `ΣJ`
+ - `SJ2`: placeholder for storing data when calculating `ΣJ`
+ - `SJ3`: placeholder for storing data when calculating `ΣJ`
+
 """
 struct LQJacobianOperator{T, V, M} <: LinearOperators.AbstractLinearOperator{T}
     Jac::M        # column of Jacobian block matrix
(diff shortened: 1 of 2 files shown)


src/DynamicNLPModels.jl
```diff
@@ -2207,14 +2197,72 @@ end
...
+    fill!(Jac.SJ2, T(0))
+    fill!(Jac.SJ3, T(0))
+end
+
+"""
+    add_jtsj!(H::M, Jac::LQJacobianOperator{T, V, M}, Σ::V)
+
+Generates `Jac' Σ Jac` and adds it to the matrix `H`.
+"""
+function add_jtsj!(
```
frapac, line 2234: I would suggest we update the signature of `add_jtsj!` to: 
```julia 
add_jtsj!(H::M, Jac::LQJacobianOperator{T, V, M}, Σ::V, alpha::Number, beta::Number)
```
That way, we can choose to update `H` inplace `beta=1` or to overwrite `H` (`beta=0`)

test/runtests.jl
```diff
@@ -1,6 +1,69 @@
 using Test, DynamicNLPModels, MadNLP, Random, JuMP, LinearAlgebra, SparseArrays, CUDA
 include("sparse_lq_test.jl")
 
+function tril_to_full!(dense::Matrix{T}) where T
+    for i=1:size(dense,1)
+        Threads.@threads for j=i:size(dense,2)
```
frapac: Can you remove `Threads.@threads` in that function?

frapac (approved):
Nice! The PR looks good to merge. 
The next step is to try to figure out how we can exploit parallelism inside `add_jtsj!`

src/DynamicNLPModels.jl
```diff
@@ -2204,17 +2194,90 @@ function Base.isreal(
...
-    fill!(Jac.JB3, T(0))
+    fill!(Jac.SJ1, T(0))
+    fill!(Jac.SJ2, T(0))
+    fill!(Jac.SJ3, T(0))
+end
+
+"""
+    add_jtsj!(H::M, Jac::LQJacobianOperator{T, V, M}, Σ::V)
+
+Generates `Jac' Σ Jac` and adds it to the matrix `H`.
```
sshin23, line 2230: also add description of what keyword arguments do

src/DynamicNLPModels.jl
```diff
@@ -2204,17 +2194,90 @@ function Base.isreal(
...
+"""
+function add_jtsj!(
+    H::M,
+    Jac::LQJacobianOperator{T, V, M},
+    Σ::V,
+    alpha::Number = 1,
+    beta::Number = 1
```
sshin23, line 2239: nice

src/DynamicNLPModels.jl
```diff
@@ -369,16 +369,10 @@ Attributes
...
- - `J1B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
- - `J2B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
- - `J3B`  : matrix for storing multiplicaitons when doing `J^T Sigma J`
+ - `SJ1`: placeholder for storing data when calculating `ΣJ`
+ - `SJ2`: placeholder for storing data when calculating `ΣJ`
+ - `SJ3`: placeholder for storing data when calculating `ΣJ`
+
 """
 struct LQJacobianOperator{T, V, M} <: LinearOperators.AbstractLinearOperator{T}
     Jac::M        # column of Jacobian block matrix
```
sshin23: Now I find this naming a bit confusing; `Jac` sounds like it's full Jacobian. Can we rename this thing? I suggest `ctrbm` https://en.wikipedia.org/wiki/Controllability but open for any other suggestion

sshin23 (approved):
well done, @dlcole3

sshin23:
one last thing, `jac_op` should be extended 
https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/51841a1f092bd2fad0bfd1b9c7ec74c3cc94ace0/src/nlp/api.jl#L599-L605

src/DynamicNLPModels.jl
dlcole3: @sshin23 @frapac  What do you think about `truncated_Jac` or `implicit_Jac`. My initial thought is that this matrix is for simplifying storage rather than changing controlibility, since a portion of the Jacobian matrix contains variable constraints. `ctrbm` seems to me like it refers to the `block_B` matrix rather than the Jacobian, but I'm not positive.

sshin23: `truncated_jac` sounds right to me

PR #39, opened 2022-07-25
Changed `LQJacobianOperator` to store jacobian as 3 tensors and added support for `CUDA.jl`

dlcole3:
- The `LQJacobianOperator` now stores the implicit Jacobian as 3 tensors rather than 1 matrix (this makes the block matrices contiguous in memory). I also updated the `mul!` functions and `add_jtsj!` to work with these tensors (`mul!` functions now take 2 for loops rather than 1). 
 - The tensors use a new function `_dnlp_unsafe_wrap` to access the data from the tensors (results in less allocations than calling `view` within `LinearAlgebra.mul!`).
 - Added CUDA as a dependency. In the future, we may want to make the GPU functions a separate package/extension
 - Added new `mul!` and `add_jtsj!` functions that work with `CuArrays`. These functions are defined with multiple dispatch, and the functions for calculating $Jx$ and $J^T x$ use `gemm_strided_batched!`. This could be optimized more in the future
 - `add_jtsj!` with `CuArrays` is not close to optimal. Unfortunately, `mul!(C, A, B, alpha, beta)` does not work with `CuArrays` when `C` is accessed with a `view`, so I had to define a new storage matrix, `H_sub_block`, within `LQJacobianOperator`. I expect this will change in the future when I update this function with `CUBLAS` functions. 
 - Added tests for making sure that `add_jtsj!` and `mul!` using the GPU give the correct answer.

### DIFF
--- a/src/DynamicNLPModels.jl
+++ b/src/DynamicNLPModels.jl
@@ -5,6 +5,7 @@ import QuadraticModels
 import LinearAlgebra
 import SparseArrays
 import LinearOperators
+import CUDA
 
 import SparseArrays: SparseMatrixCSC
 
@@ -146,20 +147,20 @@ function LQDynamicData(
         error("size of Q is not consistent with length of s0")
     end
 
-    if _cmp_arr(<, sl, su)
+    if !all(sl .<= su)
         error("lower bound(s) on s is > upper bound(s)")
     end
-    if _cmp_arr(<, ul, uu)
+    if !all(ul .<= uu)
         error("lower bound(s) on u is > upper bound(s)")
     end
-    if _cmp_arr(>=, s0, sl) || _cmp_arr(<=, s0, su)
+    if !all(sl .<= s0) || !all(s0 .<= su)
         error("s0 is not within the given upper and lower bounds")
     end
 
     if size(E, 1) != size(F, 1)
         error("E and F have different numbers of rows")
     end
-    if _cmp_arr(<, gl, gu)
+    if !all(gl .<= gu)
         error("lower bound(s) on Es + Fu is > upper bound(s)")
     end
     if size(E, 2) != size(Q, 1)
@@ -363,7 +364,9 @@ to `LinearAlgebra.mul!()`.
 
 ---
 Attributes
- - `truncated_jac`: Matrix of first `nu` columns of the Jacobian
+ - `truncated_jac1`: Matrix of first `nu` columns of the Jacobian corresponding to Ax + Bu constraints
+ - `truncated_jac2`: Matrix of first `nu` columns of the Jacobian corresponding to state variable bounds
+ - `truncated_jac3`: Matrix of first `nu` columns of the Jacobian corresponding to input variable bounds
  - `N`  : number of time steps
  - `nu` : number of inputs
  - `nc` : number of algebraic constraints of the form gl <= Es + Fu <= gu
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -469,6 +570,15 @@ lq_dense_imp = DenseLQDynamicModel(dnlp; implicit = true)
...
 
+# Test mul! operators with CUDA
+if CUDA.has_cuda_gpu()
+    dnlp_cuda     = dynamic_data_to_CUDA(dnlp)
+    lq_dense_cuda = DenseLQDynamicModel(dnlp_cuda; implicit=true)
+
+    test_mul(lq_dense, lq_dense_cuda; cuda=true)
+    test_add_jtsj(lq_dense, lq_dense_cuda; cuda=true)
+end
+
```
sshin23, line 581: can we do the tests with for a loop?

sshin23 (approved):
Nice work, @dlcole3

src/DynamicNLPModels.jl
```diff
@@ -372,20 +375,33 @@ Attributes
...
-    truncated_jac::M  # column of Jacobian block matrix
-    N::Int            # number of time steps
-    nu::Int           # number of inputs
-    nc::Int           # number of inequality constraints
-    nsc::Int          # number of state variables that are constrained
-    nuc::Int          # number of input variables that are constrained
-
-    # Storage matices for building J^TΣJ
+struct LQJacobianOperator{T, M, A} <: LinearOperators.AbstractLinearOperator{T}
+    # TODO: remove V from operator type
```
frapac: Is this comment still relevant?

src/DynamicNLPModels.jl
```diff
@@ -976,7 +991,10 @@ function _build_implicit_dense_lq_dynamic_model(dnlp::LQDynamicData{T,V,M,MK}) w
     num_real_bounds_s = sum(bool_vec_s)
 
     G          = _init_similar(Q, nc * N, nu, T)
-    Jac        = _init_similar(Q, nc * N + num_real_bounds_s * N, nu, T)
+    Jac1       = _init_similar(Q, nc, nu, N, T)
+    Jac2       = _init_similar(Q, num_real_bounds_s, nu, N, T)
+    Jac3       = _init_similar(Q, 0, nu, N, T)
+    #Jac        = _init_similar(Q, nc * N + num_real_bounds_s * N, nu, T)
```
frapac: Maybe remove dead comment?

src/DynamicNLPModels.jl
```diff
@@ -1261,7 +1307,7 @@ function _build_block_matrices(
     A::M, B::M, K, N
 ) where {T, M <: AbstractMatrix{T}}
 
-    ns = size(A, 2)
+ns = size(A, 2)
```
frapac: indentation issue?

src/DynamicNLPModels.jl
```diff
@@ -1369,11 +1408,11 @@ function _build_H_blocks(Q, R, block_A::M, block_B::M, S, Qf, K, s0, N) where {T
 
     LinearAlgebra.axpy!(1.0, Q, quad_term)
     LinearAlgebra.axpy!(1.0, SK, quad_term)
-    LinearAlgebra.axpy!(1.0, SK', quad_term)
+    quad_term .+= SK'
```
frapac, line 1413: Can we add a comment here to explain why we are using the broadcast operator for the transpose operation?

src/DynamicNLPModels.jl
```diff
@@ -2091,61 +2128,177 @@ function _init_similar(mat, dim1::Number, T=eltype(mat))
...
+        x2_view = view(x2, :, :, i:N)
+        x3_view = view(x3, :, :, i:N)
+
+        J1_view = view(J1, :, :, 1:(N - i + 1))
+        J2_view = view(J2, :, :, 1:(N - i + 1))
+        J3_view = view(J3, :, :, 1:(N - i + 1))
+
+        y1_view = view(y1, :, :, i:N)
+
+        CUDA.CUBLAS.gemm_strided_batched!('N', 'N', 1, J1_view, y1_view, 1, x1_view)
```
frapac: Can we replace `CUDA.CUBLAS` by just `CUBLAS`?

src/DynamicNLPModels.jl
```diff
@@ -2091,61 +2128,177 @@ function _init_similar(mat, dim1::Number, T=eltype(mat))
...
+        CUDA.CUBLAS.gemm_strided_batched!('N', 'N', 1, J2_view, y1_view, 1, x2_view)
+        CUDA.CUBLAS.gemm_strided_batched!('N', 'N', 1, J3_view, y1_view, 1, x3_view)
 
-        LinearAlgebra.mul!(view(y, (1 + (i - 1) * nc):(N * nc)), sub_B1, sub_x, 1, 1)
-        LinearAlgebra.mul!(view(y, (1 + nc * N + (i - 1) * nsc):((nc + nsc) * N)), sub_B2, sub_x, 1, 1)
-        LinearAlgebra.mul!(view(y, (1 + (nc + nsc) * N + (i - 1) * nuc):((nc + nsc + nuc) * N)), sub_B3, sub_x, 1, 1)
     end
+
+    view(x, 1:(nc * N)) .= reshape(x1, nc * N)
```
frapac: Do we need a view on the left-hand-side?

src/DynamicNLPModels.jl
```diff
@@ -2091,61 +2128,177 @@ function _init_similar(mat, dim1::Number, T=eltype(mat))
...
 
+
+function LinearAlgebra.mul!(
+    y::V,
+    Jac::LinearOperators.AdjointLinearOperator{T, LQJacobianOperator{T, M, A}},
+    x::V
+) where {T, V <: CUDA.CuArray{T, 1, CUDA.Mem.DeviceBuffer}, M <: AbstractMatrix{T}, A <: AbstractArray{T}}
+    fill!(y, zero(T))
+
+    J1  = get_jacobian(Jac).truncated_jac1
```
frapac: Can we use instead? 
```julia
jac_op = get_jacobian(Jac)
J1 = jac_op.truncated_jac1
...
```

frapac (approved):
It overall looks good to me, apart of a few minor comments

src/DynamicNLPModels.jl
dlcole3: Nope! I will remove it :)

PR #42, opened 2022-08-03
Added support to efficiently build `SparseLQDynamicModel` from sparse `A, B, Q, R, K, S`

### LINKED ISSUE #33: Add support for sparse `Q, R, S` matrix in building `SparseLQDynamicModel`
The current formulation for building a `SparseLQDynamicModel` is inefficient if the `Q`, `R`, or `S` matrices are sparse. Would like to add support to handle these matrices efficiently when building `H_colptr`, `H_rowval`, and `H_nzval`.

dlcole3:
Added new `_build` and `_set` functions in the `sparse.jl` source code to support sparse `A, B, Q, R, K,` and `S` matrices. Currently, when the problem becomes very large, the sparse formulation is not very fast and it takes a lot of memory allocation. This is largely because the old code treated all `A, B, Q, R, K,` and `S` matrices as being dense. Now, the Hessian and Jacobian are set using the `rowval`, `nzval`, and `colptr` attributes of `A, B, Q, R, K,` and `S`. This significantly reduces the memory required to build the `SparseLQDynamicModel`. 

For N = 50, nu = 10, ns = 2000, I got the following results using `@time`
dense LQDynamicData: 18.737979 seconds (200.89 k allocations: 9.577 GiB, 9.12% gc time)
sparse LQDynamicData: 0.671002 seconds (1.51 M allocations: 134.857 MiB, 2.67% gc time)

Also added tests to ensure that the Hessian and Jacobian resulting from the new sparse functions are the same as before.

### DIFF
--- a/src/LinearQuadratic/sparse.jl
+++ b/src/LinearQuadratic/sparse.jl
@@ -338,14 +338,14 @@ function _build_sparse_lq_dynamic_model(dnlp::LQDynamicData{T, V, M, MK}) where 
 
     nc = size(E, 1)
 
-    dropzeros!(A)
-    dropzeros!(B)
-    dropzeros!(Q)
-    dropzeros!(R)
-    dropzeros!(Qf)
-    dropzeros!(E)
-    dropzeros!(F)
-    dropzeros!(S)
+    SparseArrays.dropzeros!(A)
+    SparseArrays.dropzeros!(B)
+    SparseArrays.dropzeros!(Q)
+    SparseArrays.dropzeros!(R)
+    SparseArrays.dropzeros!(Qf)
+    SparseArrays.dropzeros!(E)
+    SparseArrays.dropzeros!(F)
+    SparseArrays.dropzeros!(S)
 
     H_colptr = zeros(Int, ns * (N + 1) + nu * N + 1)
     H_rowval = zeros(Int, length(Q.rowval) * N + length(R.rowval) * N + 2 * length(S.rowval) * N + length(Qf.rowval))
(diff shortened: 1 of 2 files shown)


dlcole3:
Note that the current formulation includes a `K` matrix in the sparse model. I now realize I don't need that, but I will fix that on a separate PR. I already had the new functions enabled with `K` when I learned I did not need to use `K` with the sparse form

src/LinearQuadratic/sparse.jl
```diff
@@ -313,6 +313,259 @@ function _build_sparse_lq_dynamic_model(
...
+
+    _set_sparse_H!(H_colptr, H_rowval, H_nzval, Q, R, N; Qf = Qf, S = S)
+
+    H = SparseArrays.SparseMatrixCSC((N + 1) * ns + nu * N, (N + 1) * ns + nu * N, H_colptr, H_rowval, H_nzval)
+
+    _set_sparse_J!(J_colptr, J_rowval, J_nzval, A, B, E, F, K, N)
+
+    J = SparseArrays.SparseMatrixCSC((nc + ns) * N, (N + 1) * ns + nu * N, J_colptr, J_rowval, J_nzval)
+
+    SparseArrays.dropzeros!(H)
```
sshin23: `dropzeros!` can be performed for `Q,R,A,B` matrices instead

sshin23 (approved):
looks good @dlcole3

test/runtests.jl
```diff
@@ -108,6 +108,17 @@ function dynamic_data_to_CUDA(dnlp::LQDynamicData)
     )
 end
 
+function test_sparse_support(lqdm)
+d = lqdm.dynamic_data
```
frapac: Indentation!

frapac (approved):
Looks good to me, except one issue with the indentation!

PR #44, opened 2022-08-05
Added functions to reset `s0` within the dense and sparse models

### LINKED ISSUE #43: Add support for udpating `DynamicData` and `LQDynamicModels` when s0 changes at sample periods
Neither the `DenseLQDynamicModel` or the `SparseLQDynamicModel` currently has support for changing `s0` at each new sample period. Need to add a function that will reset `s0` within the `LQDynamicData` and reset certain vectors in the `LQDynamicModel`. In the case of the `SparseLQDynamicModel`, we can update the variable bounds on `s0`. In the case of the `DenseLQDynamicModel`, we need to update the linear term and the constant term of the objective function, and we need to update some of the constraint bounds.

dlcole3:
After each sample period, the new initial states, `s0`, need to be reset within the model. This PR adds support for resetting that value within the dynamic data and impacted parts of the dynamic model. The function `reset_s0!` is now exported so that the user can reset this value. In addition, I added tests for this function within `runtests.jl` and I reformatted the `test/` subdirectory slightly so that all of the functions are in their own file. 

In the sparse formulation, `reset_s0!` resets `s0` within the dynamic data, and it resets the variable bounds on `s` at time 0 (thus fixing the first state to the value of `s0`. 

For the dense formulation, `reset_s0!` resets `s0` in the dynamic data, and it also calculates a new linear and constant term in the objective function. It  also changes the impacted constraint bounds for the Jacobian. Because resetting `s0` should be a quick and efficient call, I added blocks to the `DenseLQDynamicBlocks` that store the matrices needed to get the new linear and constant terms and for calculating the new constraint bounds. This also meant reconfiguring the functions `_build_H_blocks` and `_set_G_blocks!`. The data within `DenseLQDynamicBlocks` is designed to just be multiplied by `s0` to get the desired final values.

### DIFF
--- a/src/LinearQuadratic/dense.jl
+++ b/src/LinearQuadratic/dense.jl
@@ -1,4 +1,3 @@
-
 """
 DenseLQDynamicModel(dnlp::LQDynamicData; implicit = false)    -> DenseLQDynamicModel
 DenseLQDynamicModel(s0, A, B, Q, R, N; implicit = false ...) -> DenseLQDynamicModel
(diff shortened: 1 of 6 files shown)


src/LinearQuadratic/dense.jl
```diff
@@ -931,10 +968,9 @@ function _set_G_blocks!(
     G[1:nc, :] = F
 
     EB   = _init_similar(block_B, nc, nu, T)
-    As0  = _init_similar(s0, ns * (N + 1), T)
-    EAs0 = _init_similar(s0, nc, T)
+    EA   = _init_similar(block_B, nc, ns, T)
 
-    LinearAlgebra.mul!(As0, block_A, s0)
+    #LinearAlgebra.mul!(As0, block_A, s0)
```
frapac: maybe remove dead comment?

frapac (approved):
Looks good to me!

PR #45, opened 2022-08-12
Added support for constant term in the linear dynamic constraints

dlcole3:
For $x_{t + 1} = Ax_t + bu_t + w$, The vector `w` changes the constraint bounds and the linear/constant terms of the objective function. I derived the equations for these and updated the `SparseLQDynamicModel` and `DenseLQDynamicModel` to operate with `w`. I added tests to check this, and I updated the JuMP model that we used for comparison. This also required updating the `DenseLQDynamicBlocks` significantly to store the needed data to reset `s0`. 

For reference, the equation used for deriving this is $\boldsymbol{x} = \boldsymbol{A} \hat{x} + \boldsymbol{B} \boldsymbol{v} + \tilde{\boldsymbol{A}}\boldsymbol{w}$. In this case, $\tilde{\boldsymbol{A}}$ is the same format as $\boldsymbol{B}$ but with the identity matrix in place of the $B$ matrix.

### DIFF
--- a/src/LinearQuadratic/LinearQuadratic.jl
+++ b/src/LinearQuadratic/LinearQuadratic.jl
@@ -211,7 +211,7 @@ and u = [u_0^T, u_1^T, ..., u_{N-1}^T]
 - `h`  : nu(N) x ns matrix for building the linear term of the objective function. Just needs to be
 multiplied by `s0`.
 - `h01`: ns x ns matrix for building the constant term fo the objective function. This can be found by
-taking $ s_0^T h01 s_0$
+taking  s_0^T h01 s_0
 - `h02`: similar to `h01`, but one side is multiplied by `Aw` rather than by `As0`. This will just
 be multiplied by `s0` once
 - `h_constant` : linear term in the objective function that arises from `Aw`. Not a function of `s0`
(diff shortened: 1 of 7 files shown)


sshin23:
@dlcole3 can't `w` be time-varying? When we have `w`, typically we have time-variability in `w`

src/LinearQuadratic/LinearQuadratic.jl
```diff
@@ -192,35 +199,64 @@ struct SparseLQDynamicModel{T, V, M1, M2, M3, MK} <:  AbstractDynamicModel{T,V}
...
 
 See also `reset_s0!`
 """
-struct DenseLQDynamicBlocks{T, M}
-    # block_h0 = A^T((Q + KTRK + 2 * SK))A where Q, K, R, S, and A are block matrices
-    # block_h  = (QB + SKB + K^T R K B + K^T S^T B)^T A + (S + K^T R)^T A
-    # block_d  = (E + FK) A
-    # block_KA = KA
+mutable struct DenseLQDynamicBlocks{T, V, M}
+    # Aw = block_matrix_A * w (result is a Vector; block_matrix A is like block_B,
```
frapac: Can we move the comments in front of the attribute in the struct?

frapac (approved):
it overall looks good to me!

dlcole3:
> @dlcole3 can't `w` be time-varying? When we have `w`, typically we have time-variability in `w`

@sshin23 It could be time varying, and that would be a relatively simple change to make. I will update that. In the case of the application in the second example problem (oscillating masses), they are not time varying.

sshin23 (approved):
If time-varying is taking long time, feel free to merge this first and open an issue.

The change looks good

sshin23 (approved):

